PR #2, opened 2019-12-31
Update travis configuration

mingfeng:
Previously travis build was broken

no refs

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,25 +1,40 @@
-sudo: false
-language: python
-services:
-  - postgresql
+dist: xenial
 
-cache:
-  directories:
-    - "$HOME/.cache/pip"
+language: python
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -1,25 +1,40 @@
...
+  - "3.8"
+
+cache: pip
+
 services:
-  - postgresql
+   - postgresql
+
+addons:
+  postgresql: "10"
```
mhieta: Is it possible to use latest Postgres/PostGIS-releases? E.g. Postgres 12.x and PostGIS 3.0.0?

mhieta: Prebuild postgres-12-postgis-3 packages are not available under xenial. Sticking with the current configuration.

PR #3, opened 2020-01-10
Feature/traffic sign real

mhieta:
PR contains following changes:
- Add Traffic Sign Realization model
- Expose Traffic Sign Realization in REST API
- Expose Traffic Sign Realization in Django Admin UI
- Add i18n support
- Translate project to finnish


traffic_control/models.py
```diff
@@ -160,3 +176,101 @@ class Meta:
 
     def __str__(self):
         return "%s %s %s" % (self.id, self.code, self.value)
+
+
+class TrafficSignReal(models.Model):
```
ehaivala: This model represents a regular traffic sign that's physically somewhere, right? In that case I don't really see the point of having `Real` suffix in the model name. In my opinion `TrafficSign` is enough in the context of this project and that the `Real` doesn't have additional value.

traffic_control/models.py
```diff
@@ -160,3 +176,101 @@ class Meta:
...
+        _("Installation details"), max_length=254, blank=True, null=True
+    )
+    condition = models.IntegerField(choices=Condition.choices, default=Condition.GOOD)
+    decision_id = models.CharField(_("Decision id"), max_length=254)
+    size = models.CharField(max_length=1, choices=Size.choices, default=Size.MEDIUM)
+    reflection_class = models.CharField(
+        max_length=2, choices=Reflection.choices, default=Reflection.R1
+    )
+    surface_class = models.CharField(
+        max_length=6, choices=Surface.choices, default=Surface.DIRECT
```
ehaivala: Not strictly related to this PR but does `Surface.DIRECT` indicate that the sign is not curved in any way? Are these terms specified by someone else or how these are determined?

I'm fairly sure that when speaking of surfaces, "direct" is not the proper term to say that it is not curved or crooked in any way. I'd say that kind of surface is either "flat" or "straight" instead.

traffic_control/locale/fi/LC_MESSAGES/django.po
```diff
@@ -0,0 +1,331 @@
...
+msgid "Height"
+msgstr "Korkeus"
+
+#: traffic_control/models.py:109 traffic_control/models.py:204
+msgid "Direction"
+msgstr "Suunta"
+
+#: traffic_control/models.py:112 traffic_control/models.py:207
+msgid "Traffic sign code value"
+msgstr "Liikennemerkin arvo"
```
ehaivala: "Traffic sign code" is translated as "Merkkikoodi" on L163. The translation should probably match so that same term is translated the same way for each occurance.

traffic_control/locale/fi/LC_MESSAGES/django.po
```diff
@@ -0,0 +1,331 @@
...
+msgid "Plan link"
+msgstr "Suunnitelmalinkki"
+
+#: traffic_control/models.py:161 traffic_control/models.py:270
+msgid "Size"
+msgstr "Koko"
+
+#: traffic_control/models.py:164 traffic_control/models.py:273
+msgid "Reflection"
+msgstr "Kalvo"
```
ehaivala: This seems wrong. Not sure if there's some industry terminology I'm not aware of but pointing this out just in case.

traffic_control/locale/fi/LC_MESSAGES/django.po
```diff
@@ -0,0 +1,331 @@
...
+#: traffic_control/models.py:171 traffic_control/models.py:280
+msgid "Lane number"
+msgstr "Kaistanumero"
+
+#: traffic_control/models.py:172 traffic_control/models.py:281
+msgid "Lane type"
+msgstr "Kaistatyyppi"
+
+#: traffic_control/models.py:174 traffic_control/models.py:283
+msgid "LocationSpecifier"
```
ehaivala: This kind of translatable class names should probably formatted to a human friendly format. That can be done at a later date though.

traffic_control/migrations/0001_initial.py
```diff
@@ -259,6 +273,7 @@ class Migration(migrations.Migration):
                         ],
                         default=1,
                         null=True,
+                        verbose_name="LocationSpecifier",
```
ehaivala: Related to the previous comment. This should be "Location specifier" instead.

traffic_control/migrations/0001_initial.py
```diff
@@ -514,6 +541,7 @@ class Migration(migrations.Migration):
...
```
ehaivala: Same here.

traffic_control/locale/fi/LC_MESSAGES/django.po
aslani: It's industry term

traffic_control/models.py
mhieta: In this application context, it is important to separate plan and real models from each other. `Real` makes the model more explicit, thus I would prefer leaving it there.

traffic_control/models.py
mhieta: Let's change it to `Surface.FLAT`

traffic_control/models.py
ehaivala: True. Explicit is better than implicit. Let's leave this like this. 👍

traffic_control/locale/fi/LC_MESSAGES/django.po
mhieta: Let's update it to "Merkkikoodin arvo".

traffic_control/locale/fi/LC_MESSAGES/django.po
mhieta: Updated to "Location specifier"

traffic_control/migrations/0002_auto_20200109_1500.py
```diff
@@ -0,0 +1,130 @@
+# Generated by Django 3.0.1 on 2020-01-09 13:00
```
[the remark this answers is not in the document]
mhieta: True. Updated.

ehaivala:
The commits could still be cleaned up a bit. Currently there's stuff related to translations in three of the five commits. It would be nice to have all of the stuff related to translation in same commit to keep rest of the logic nicely separated from that.

On the other hand it doesn't really matter that much this early on when we're only laying out the basics for the project. It'd be nice to clean the history up a bit but I'll leave the decision up to you.

To me this looks good otherwise :+1:

PR #7, opened 2020-01-20
Add Mounts - planned datamodel.

mhieta:
Add REST API.
Add Admin UI.
Add tests.
Update TrafficSignPlan-model to have reference to MountPlan.
Update translations.


.gitignore
```diff
@@ -77,3 +77,4 @@ sftp-config.json
 target/
 var/
 venv/
+/locale/
```
ehaivala: As a side note not strictly related to the contents of this PR:

Instead of only ignoring the repository root level locale directory, we should not generate translations for the third party packages in the first place. Adding `--ignore=venv/*` to `makemessages` command that should take care of that but it's easily forgotten if it needs to be manually added every time updating translations. Should there be a basic bash script for generating the locales in future?

mhieta: Good point. We could make a bash script or would it be enough to just have that command in the README to avoid extra custom scripts?

mhieta: Done. Added bash-script for translations.

traffic_control/locale/fi/LC_MESSAGES/django.po
```diff
@@ -46,357 +46,395 @@ msgstr "Vaijeri"
...
 msgid "Traffic Sign Real"
 msgstr "Liikennemerkki (toteuma)"
 
-#: traffic_control/models.py:459
+#: traffic_control/models.py:495
 msgid "Traffic Sign Reals"
 msgstr "Liikennemerkki (toteumat)"
+
+#~ msgid "Type"
+#~ msgstr "Tyyppi"
```
ehaivala: These can be removed

mhieta: Fixed

PR #17, opened 2020-01-31
Refactor models and tests.

mhieta:
Refactor models and tests to entity-based modules.
Update migrations.
Update translations.


traffic_control/models/__init__.py
```diff
@@ -0,0 +1,7 @@
+from .barrier import *
```
ehaivala: Import only the model classes instead of using `*` imports.

mhieta: Good point. Case handled.

PR #20, opened 2020-02-07
Feature/traffic sign codes import

mhieta:
Import Traffic Sign Codes from CSV-file.
- Add Traffic Sign Code REST API (with filtering and ordering capabilities)
- Add Python script that reads CSV-file of Traffic Sign Codes and sends those to the user provided REST API

Refs LIIK-30, LIIK-49


traffic_control/management/commands/import_traffic_sign_codes.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def add_arguments(self, parser):
+        parser.add_argument("filename", help="Path to the traffic sign codes csv file")
+
+    def handle(self, *args, **options):
+        filename = options["filename"]
+        if not os.path.exists(filename):
+            raise CommandError("File {0} does not exist".format(filename))
+
+        with open(filename, mode="r", encoding="utf-8-sig") as csv_file:
```
aslani: I sort of don't like what's going on here. First of all if we are running this command "locally" aka in the environment that we are running this code, we don't need to use the API client to do the POSTing. At least since receiving these objects are trivial task and there's no special stuff with them. Hmm. Also I don't like this:
`user = User.objects.create_superuser(
                username=username, password=password, email="testadmin@example.com"
            )`
It's just super hacky way to do things... and if we wanted to do it this way aka manage.py command with API client, just read the credentials for the user from environment variables.

Sad to say, but I think this approach is most likely not the way to go - even when the code itself is actually totally fine! Hopefully this does not sound too harsh :)

traffic_control/views.py
```diff
@@ -103,3 +110,29 @@ class RoadMarkingPlanViewSet(TrafficControlViewSet):
...
+
+class TrafficSignCodeViewSet(ModelViewSet):
+    filter_backends = [DjangoFilterBackend, OrderingFilter]
+    ordering_fields = "__all__"
+    ordering = ["code"]
+    permission_classes = [IsAdminUser]
+    serializer_class = TrafficSignCodeSerializer
+    queryset = TrafficSignCode.objects.all()
+
+    def create(self, request, *args, **kwargs):
```
aslani: I'm not totally convinced that this is actually even needed. You could just have the "validation" in the model's clean method - if it fails then it should return 400 as well. In this case in the code has to be unique, then just have unique = true in the model or something like that and you don't even have to do the custom clean in the model.

PR #21, opened 2020-02-10
Feature/models filtering

Jogha:
Add filtering for the rest of the views. Includes little refactoring and formatting.

Refs: LIIK-40, LIIK-41, LIIK-42, LIIK-43, LIIK-44, LIIK-45


traffic_control/tests/test_barrier_api.py
```diff
@@ -1,11 +1,37 @@
...
+    TrafficControlAPIBaseTestCase,
+)
+
+
+@pytest.mark.django_db
+@pytest.mark.parametrize(
+    "location,query_location,expected",
+    [*point_location_test_data, *line_location_test_data],
+)
+def test_filter_barrier_plans_location(location, query_location, expected):
```
mhieta: About code readability in `{"location": query_location.ewkt}`. It might be better to use `location_query` or similar to make code more readable.

traffic_control/tests/test_barrier_api.py
```diff
@@ -1,11 +1,37 @@
...
+    """
+    api_client = get_api_client()
+
+    get_barrier_plan(location)
+    response = api_client.get(
+        reverse("api:barrierplan-list"), {"location": query_location.ewkt}
+    )
+
+    assert response.status_code == status.HTTP_200_OK
+    assert response.data.get("count") == expected
```
mhieta: Maybe better to validate also the response contents somehow, like check if it's empty list or none when geometry query does not return anything.

Jogha: Added ID checking for the response, which should cover everything in my opinion.

traffic_control/tests/test_barrier_api.py
Jogha: Changed as requested.

PR #31, opened 2020-02-24
Sentry integration

Jogha:
Raven is currently phasing out in favor of Sentry SDK, which also
supports newest versions of Python and Django. Our project will be
updated to Django 3 later, so we want to use packages that support
that.

- Remove Raven
- Add Sentry SDK
- Configure Sentry SDK in settings
- Add debug path for sentry

Refs: LIIK-54

### DIFF
--- a/city-infrastructure-platform/settings.py
+++ b/city-infrastructure-platform/settings.py
@@ -11,7 +11,6 @@ https://docs.djangoproject.com/en/2.2/ref/settings/
 import os
 
 import environ
-import raven
 from django.utils.translation import gettext_lazy as _  # NOQA
 
 # Set up .env file
@@ -41,11 +40,6 @@ env = environ.Env(
 if os.path.exists(env_file):
     env.read_env(env_file)
 
-try:
-    version = raven.fetch_git_sha(checkout_dir())
-except Exception:
-    version = None
-
 # General settings
 DEBUG = env.bool("DEBUG")
 TIER = env.str("TIER")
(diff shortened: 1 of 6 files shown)


city-infrastructure-platform/settings.py
```diff
@@ -162,5 +164,10 @@
 # Azure CLIENT_IP middleware
 AZURE_DEPLOYMENT = env.bool("AZURE_DEPLOYMENT")
 
+# Sentry-SDK
+SENTRY_DSN = env.str("SENTRY_DSN")
+if SENTRY_DSN:
+    sentry_sdk.init(dsn=SENTRY_DSN, integrations=[DjangoIntegration()])
```
ehaivala: What's your opinion on having the Git commit hash here? It could be passed as `release` keyword argument to the `init` function.

Is the release/commit information so useful that we want to keep it?

Jogha: As discussed in Slack, I'll add this release information here

city-infrastructure-platform/utils.py
```diff
@@ -0,0 +1,31 @@
...
+        return out
+
+    try:
+        out = _minimal_ext_cmd(["git", "rev-parse", "HEAD"])
+        git_revision = out.strip().decode("ascii")
+    except (subprocess.SubprocessError, OSError):
+        git_revision = "Unknown"
+
+    if not git_revision:
+        # this shouldn't happen but apparently can (see gh-8512)
```
ehaivala: You can leave out the part in parenthesis.

city-infrastructure-platform/utils.py
```diff
@@ -0,0 +1,31 @@
+import os
```
ehaivala: We're working with an open source project that is borrowing code from Numpy, another open source project, so we definitely should include the licence Numpy is using at the start of this file.

Jogha: Good point! :+1:

PR #32, opened 2020-02-25
Feature/traffic sign reals import

mhieta:
Add script for importing Traffic Sign Reals from SHP-file.

Script reads the provided Traffic Sign Real shapefile's first layer, transforms features into suitable format and sends them one by one to REST API `/api/traffic-sign-reals/` .
Script in intended to be run outside of the Django project and requires a working GDAL-installation in order to run it.

API updates:
- Make Traffic Sign Code non-mandatory for Traffic Sign Real
- Add `legacy_code` and `legacy_description` to Traffic Sign Code model
- Add `legacy_code` and `scanned_at` to Traffic Sign Real model
- Automatically link Traffic Sign Code to TrafficSign Real when code does not exist but legacy code exists
- Add `source_id` and `source_name` to Traffic Sign Real and Traffic Sign Plan models
- Change `mount_type` to be CharField in Traffic Sign models
- Add `mount_type_fi` for finnish version of the mount type to Traffic Sign models
- Add fields to Django Admin
- Add migrations
- Update translations

Refs LIIK-32, LIIK-57


traffic_control/utils/send_traffic_sign_reals.py
```diff
@@ -0,0 +1,77 @@
...
+    print("Processing Feature ID: {0}".format(feature.GetFID()))
+    feature_json = feature.ExportToJson(as_object=True)
+    geometry = feature_json["geometry"]
+    point = ogr.Geometry(ogr.wkbPoint)
+    point.AddPoint_2D(geometry["coordinates"][0], geometry["coordinates"][1])
+    properties = feature_json["properties"]
+    data = {
+        "location": "SRID={0};{1}".format(SOURCE_SRID, point.ExportToWkt()),
+        "height": round(geometry["coordinates"][2], 6),
+        "legacy_code": properties["type"].strip() if properties["type"] else None,
```
ehaivala: The `properties["x"].strip() if properties["x"] else None` is repeated so many times that it would make sense to have a function to do it.

traffic_control/utils/send_traffic_sign_reals.py
```diff
@@ -0,0 +1,77 @@
...
+        "txt": properties["text"].strip() if properties["text"] else None,
+        "mount_type": properties["mount_type"].strip()
+        if properties["mount_type"]
+        else None,
+        "direction": properties["direction"].strip()
+        if properties["direction"]
+        else None,
+        "scanned_at": properties["date"],
```
ehaivala: Does this date exist always? The script will raise a `KeyError` here if it's missing.

traffic_control/utils/send_traffic_sign_reals.py
```diff
@@ -0,0 +1,77 @@
...
+if not os.path.exists(filename):
+    raise Exception("File {0} does not exist".format(filename))
+
+file = ogr.Open(filename)
+layer = file.GetLayer(0)
+print("SHP-file has {0} features.".format(layer.GetFeatureCount()))
+auth = HTTPBasicAuth(args.username, args.password)
+counter = 0
+layer.ResetReading()
+for feature in layer:
```
ehaivala: It's a good practice to add an empty line before a bigger block like this `for`.

traffic_control/models/traffic_sign.py
```diff
@@ -324,6 +338,20 @@ class Meta:
     def __str__(self):
         return "%s %s %s" % (self.id, self.code, self.value)
 
+    def save(self, *args, **kwargs):
+        if not self.code:
+            try:
```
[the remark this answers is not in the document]
mhieta: If the code exists, then we should not do anything. If the code does not exist then we try to retrieve the code using legacy_code

traffic_control/utils/send_traffic_sign_reals.py
mhieta: Will add function

traffic_control/utils/send_traffic_sign_reals.py
mhieta: Date always exists in the data, but better account for nulls

traffic_control/utils/send_traffic_sign_reals.py
mhieta: Will do

traffic_control/models/traffic_sign.py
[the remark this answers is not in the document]
ehaivala: If there's only one code then `TFS.objects.get(...)` and `TFS.objects.filter(...).order_by(...).first()` will have the same result. If there's no TrafficSignCode that matches the filtering then the value will be set to `None` as in the first except block.

I'm still fairly sure that `TrafficSignCode.objects.filter(legacy_code=self.legacy_code).order_by("code").first()` will cover all the cases in this try-except.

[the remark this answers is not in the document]
mhieta: Indeed this is the case, fixed now

PR #33, opened 2020-02-26
Auto-generate mount type finnish name

mhieta:
Auto-generate mount type finnish name content from mount type data during the Traffic Sign Real import.
Simplify existing dict retrieval code.

Refs: LIIK-64

### DIFF
--- a/traffic_control/utils/send_traffic_sign_reals.py
+++ b/traffic_control/utils/send_traffic_sign_reals.py
@@ -8,11 +8,14 @@ from requests.auth import HTTPBasicAuth
 SOURCE_SRID = 3879
 SOURCE_NAME = "BLOM streetview2019"
 OWNER = "Helsingin kaupunki"
-
-
-def strip_or_none(prop):
-    return prop.strip() if prop else None  # noqa
-
+MOUNT_TYPES_EN_FI = {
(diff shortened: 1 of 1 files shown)


traffic_control/utils/send_traffic_sign_reals.py
```diff
@@ -67,18 +70,20 @@ def strip_or_none(prop):
...
+    properties = feature_json.get("properties")
     data = {
         "location": "SRID={0};{1}".format(SOURCE_SRID, point.ExportToWkt()),
-        "height": round(geometry["coordinates"][2], 6),
-        "legacy_code": strip_or_none(properties["type"]),
-        "txt": strip_or_none(properties["text"]),
-        "mount_type": strip_or_none(properties["mount_type"]),
-        "direction": strip_or_none(properties["direction"]),
-        "scanned_at": strip_or_none(properties["date"]),
+        "legacy_code": properties.get("type", None),
```
Jogha: Get will return None already as a default value, if key is not found. It's up to you if you want to change or not  :smile: https://docs.python.org/3/library/stdtypes.html#dict.get

mhieta: Original value was empty string as default value there, so easy to miss. Will definitely change that. Good catch!

PR #40, opened 2020-03-19
Enable 3d Point editing for TrafficSignPlan and TrafficeSignReal admins

mingfeng:
Using a 2d map for xy coordinates editing and a float field for z coordinate editing.
We need to specify the list of fields in order to group the xy coordinate and z
coordinate field together.

Refs: LIIK-77

admin-3d

### DIFF
--- a/traffic_control/admin.py
+++ b/traffic_control/admin.py
@@ -4,6 +4,8 @@ from django.http import HttpResponseRedirect
 from django.urls import reverse
 from django.utils.translation import gettext_lazy as _
 
+from .forms import TrafficSignPlanModelForm, TrafficSignRealModelForm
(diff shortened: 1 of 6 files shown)


mingfeng:
@mhieta Thanks for the translation.  About the branch renaming, I think it may not worth the effort now  after pushing the PR as it requires delete old branch, create new one and remake PR if I'm not mistaken. Also if we want to make PR purpose more explicit, would it be better to  follow some conventions like this one: https://www.conventionalcommits.org/en/v1.0.0/ ?

mhieta:
In the future, we can adjust / agree more closely those PR commit rules but it needs more evaluation. Fot this PR, it's ok for now to leave it like that.

PR #54, opened 2020-03-30
Enable soft delete for REST APIs and Admin UI

mingfeng:
This PR introduced the necessary changes to enable soft delete in REST API and Admin UI including:

- When entity is deleted through Admin UI, only mark it as deleted instead of actually deleting it. This means updating deleted_by and deleted_at fields from the entity
- Exclude / do not show soft-deleted entities in the Admin UI
- Exclude soft-deleted entities from the REST API. Return status 404 in case of GET / PUT / DELETE

Refs: LIIK-85


traffic_control/tests/test_mount_api.py
```diff
@@ -430,9 +442,21 @@ def test_delete_mount_real_detail(self):
         self.assertEqual(MountReal.objects.count(), 1)
         deleted_mount_real = MountReal.objects.get(id=str(mount_real.id))
         self.assertEqual(deleted_mount_real.id, mount_real.id)
+        self.assertFalse(deleted_mount_real.is_active)
         self.assertEqual(deleted_mount_real.deleted_by, self.user)
         self.assertTrue(deleted_mount_real.deleted_at)
 
+    def test_get_deleted_mount_real_reaturns_not_found(self):
```
mhieta: Typo?

mingfeng: Good catch

PR #55, opened 2020-03-31
Add django-helusers integration

mhieta:
This PR adds django-helusers / tunnistamo integration to project.

The Admin-login procedure it currently also adjusted to work with Tunnistamo with minor manual modifications to the auto-created user object, but the actual authentication provider is disabled for now because we don't have defined the roles / groups yet in the platform level.

Refs #LIIK-89


setup.cfg
```diff
@@ -42,9 +42,10 @@ line_length = 88
 ignore=D100,D104,D105,D200,D203,D400
 
 [prequ]
+annotate = yes
 requirements =
     django-environ
-	Django~=2.2.9
+	Django~=2.2.12
```
Jogha, line 48: I feel like this django-helusers commit should be split up into multiple commits. Now there is django updates, package additions, migrations and settings all in the same commit. Would be much nicer to have atleast the django update and prequ stuff on its own.

city-infrastructure-platform/settings.py
```diff
@@ -61,7 +61,9 @@
 
 # Application definition
 DJANGO_APPS = [
-    "django.contrib.admin",
+    "helusers",
+    "social_django",
```
Jogha, line 65: In the previous commit you added this, but now you are moving it. This makes the history a bit harder to read in the future, so why not do the line change in the earlier commit?

city-infrastructure-platform/settings.py
```diff
@@ -174,10 +174,17 @@
...
+        "rest_framework.authentication.SessionAuthentication",
+    ],
+    "DEFAULT_PERMISSION_CLASSES": ["rest_framework.permissions.IsAuthenticated"],
     "DEFAULT_FILTER_BACKENDS": ["django_filters.rest_framework.DjangoFilterBackend"],
-    "PAGE_SIZE": 10,
-    "DEFAULT_AUTHENTICATION_CLASSES": ("helusers.oidc.ApiTokenAuthentication",),
+    "DEFAULT_PAGINATION_CLASS": "rest_framework.pagination.LimitOffsetPagination",
+    "PAGE_SIZE": 30,
+    "DEFAULT_RENDERER_CLASSES": ["rest_framework.renderers.JSONRenderer"],
+    "DEFAULT_SCHEMA_CLASS": "rest_framework.schemas.coreapi.AutoSchema",
```
Jogha, line 187: Should these be combined with earlier commits/on its own/or some combination? Also some explanations why these changes are done would not hurt, because the referenced ticket is tunnistamo-integration and now we are changing DRF default settings that are not related to authentication.

city-infrastructure-platform/settings.py
mhieta: It was actually removed and added again in the same commit, because helusers-packages needs to be in introduced in certain order relative to other Django-packages in order to work properly. But I will fix all these helusers-imports to be in a single commit.

city-infrastructure-platform/settings.py
mhieta: Correct, I'll make a separate commit for this.

setup.cfg
mhieta: Good point, I'll change this.

PR #65, opened 2020-04-21
Update Swagger documentation

mhieta:
Update Swagger documentation with following topics:
- Utilize schema validation
- Simplify versioning
- Add general info related to the overall API-usage
- Add geometry filter usage information
- Add brief descriptions for each entity

Also reorganize urlpatterns so that only Admin UI goes under i18n.

Refs #LIIK-68

### DIFF
--- a/city-infrastructure-platform/settings.py
+++ b/city-infrastructure-platform/settings.py
@@ -91,7 +91,7 @@ AUTH_USER_MODEL = "users.User"
 LOGIN_REDIRECT_URL = "/admin/"
 LOGOUT_REDIRECT_URL = "/admin/login/"
 SOCIAL_AUTH_TUNNISTAMO_AUTH_EXTRA_ARGUMENTS = {"ui_locales": "fi"}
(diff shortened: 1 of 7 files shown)


city-infrastructure-platform/settings.py
```diff
@@ -91,7 +91,7 @@
 LOGIN_REDIRECT_URL = "/admin/"
 LOGOUT_REDIRECT_URL = "/admin/login/"
 SOCIAL_AUTH_TUNNISTAMO_AUTH_EXTRA_ARGUMENTS = {"ui_locales": "fi"}
-WAGTAIL_SITE_NAME = _("City Infrastructure Platform Administration")
+WAGTAIL_SITE_NAME = _("City Infrastructure Platform")
```
mingfeng, line 94: I might miss something, but are we using WAGTAIL somewhere?

mhieta: No, we are not using Wagtail. But according to source code, that is the only way to localize Admin UI front-page title without completely rewriting it on a view/template basis.

PR #70, opened 2020-05-20
Plan model

ehaivala:
Add a top level `Plan` model that all other existing plan models have a foreign key relation to.

Refs LIIK-17


ehaivala:
Screenshots:

Admin `set_plan_relations_view`:
image

Autogenerated area from 5 points:
image

PR #72, opened 2020-05-25
Lane enums

ehaivala:
Update `LaneType` enum and introduce new `LaneNumber` enum. Replace existing `lane_number` char fields from various models with enum field using the new `LaneNumber` enum.

Refs LIIK-114


traffic_control/models/common.py
```diff
@@ -89,6 +89,66 @@ class Labels:
...
+    LIGHT = 20
+    PEDESTRIAN = 21
+    BIKE = 22
+
+    class Labels:
+        MAIN = _("Main lane")
+        FAST = _("Fast lane")
+        TURN_RIGHT = _("Turn right lane")
+        TURN_LEFT = _("Turn left lane")
+        ADDITIONAL = _("Additional straight lane")
```
ehaivala, line 113: I'm not sure if `ADDITIONAL` is a good name for this. I'm open for suggestions if there's something more suitable for this.

mhieta: `AUXILIARY` might be a better term for this.

PR #75, opened 2020-05-29
Add base attribute to MountPlan and MountReal

ehaivala:
A clearer separation between mount type and the mount base is needed.
Rename type attribute to mount_type and add new base attribute. Update
tests and translations.

Refs LIIK-127


ehaivala:
The status information of the other Travis build seems to be stuck somewhere. It's displayed as passed on Travis' side so I'll merge this.

PR #94, opened 2020-06-30
Feature: Additional signs

ehaivala:
Remove old TrafficSign parent relation to itself and replace it with separate AdditionalSign models.

AdditionalSigns are special because one physical sign may contain information of multiple signs so there needs to be a separate model for the sign content. So `AdditionalSignPlan` and `AdditionalSignReal` can be thought as the representation of the physical sign (with attributes like location, color, and installation_date) and `AdditionalSignContentPlan` and `AdditionalSignContentReal` can be thought as whatever information the physical sign has printed on it (such as parking allowed on saturdays between 10-18).

Refs LIIK-115


ehaivala:
Fixed the import order issue Travis complained about.

PR #97, opened 2020-07-03
Add management command to import vaisala traffic sign data

mingfeng:
Vaisala data is a relative big data set and takes a while to import all the data. The command will report step messages after 1000 rows are imported.

Refs: LIIK-156


traffic_control/management/commands/import_traffic_signs_vaisala.py
```diff
@@ -0,0 +1,63 @@
...
+            raise CommandError(f"File {filename} does not exist")
+
+        self.stdout.write("Importing vaisala traffic signs...")
+        count = 0
+        user = get_system_user()
+        with open(filename, encoding="utf-8-sig") as f:
+            csv_reader = csv.DictReader(f, delimiter=",")
+            for row in csv_reader:
+                location = Point(
+                    float(row["longitude"]), float(row["latitude"]), 0, srid=SOURCE_SRID
```
ehaivala, line 38: Doesn't the `Point` need to be transformed from the source SRID to the SRID the platform uses?

mingfeng: Yeah, it's done in next line

ehaivala: TIL that apparently I'm blind. 🤦‍♂️

PR #116, opened 2020-07-17
Admin usability updates

ehaivala:
- Verify that TrafficSign models have has_additional_signs information visible in the admin list view
- Remove installation attributes from AdditionalSignPlan as those should only exist on reals
- Organize admin views for consistency and usability
- Show Enum logical value in UI select inputs after the verbose label
- Improve AdditionalSign list views
- Fix TrafficSignReal search fields

Refs LIIK-164


traffic_control/models/additional_sign.py
```diff
@@ -195,6 +180,22 @@ class AdditionalSignReal(AbstractAdditionalSign):
...
+    installation_date = models.DateField(_("Installation date"), blank=True, null=True)
+    installation_status = EnumField(
+        InstallationStatus,
+        verbose_name=_("Installation status"),
+        max_length=10,
+        default=InstallationStatus.IN_USE,
+        blank=True,
+        null=True,
+    )
+    installed_by = models.CharField(_("Installed by"), max_length=256, blank=True)
```
aslani, line 198: So what's up here with 254 and then 256 max_length? I feel it really does not matter, but I guess we could stick to one? :D /very minor nitpick

traffic_control/admin/traffic_light.py
```diff
@@ -24,11 +29,60 @@ class TrafficLightPlanFileInline(admin.TabularInline):
...
-    SoftDeleteAdminMixin, UserStampedAdminMixin, admin.OSMGeoAdmin, AuditLogHistoryAdmin
+    EnumChoiceValueDisplayAdminMixin,
+    SoftDeleteAdminMixin,
+    UserStampedAdminMixin,
+    admin.OSMGeoAdmin,
+    AuditLogHistoryAdmin,
 ):
     default_lon = HELSINKI_LONGITUDE
     default_lat = HELSINKI_LATITUDE
     default_zoom = 12
```
aslani, line 40: I've said this before, but these three feel more like settings to me. It's very minor thing over all and does not affect the functionality of course. Just pointing it out here.

traffic_control/models/additional_sign.py
ehaivala: The `installation_details` has been added after the other fields which have max length of 256. I don't know what's the reasoning behind 254. I just moved it over as-is. 😄

PR #119, opened 2020-07-28
Feature: Operational area API permission

ehaivala:
Implement API permissions based on user or user group operational area.

Implement `OperationalArea` and `GroupOperationalArea` models. `OperationalArea` is used to define the geographic operational area and `GroupOperationalArea` is used to link `OperationalArea` instances to Django user groups.

Add API object level permission class and override `TrafficControlViewSet.create` method with permission checks as well because the object level permissions do not apply when creating new instances.

Add tests, admin views and update translations as usual.

Refs LIIK-106


ehaivala:
Don't merge yet. There's a small issue still with the permission checks this that is not covered by the tests.

ehaivala:
Should be ok now. Needed to revise the create permission checks for geojson.

ehaivala:
That message is used in tests so that the tests don't fail if the language is changed in the settings. The message we're testing against is not written by us so we shouldn't override the translation either.

mhieta:
> That message is used in tests so that the tests don't fail if the language is changed in the settings. The message we're testing against is not written by us so we shouldn't override the translation either.

Alright, then this is fine.

PR #126, opened 2020-08-19
Allow users to authenticate to REST API with Token

mhieta:
admin-token-auth

Enable standard DRF TokenAuthentication mechanism.
Tokens can be created for existing users through Django Admin UI.

Refs LIIK-171


mingfeng (approved):
LGTM :+1:

PR #129, opened 2020-08-24
Add Operational Area REST API endpoint

mingfeng:
Operational areas are used to restrict the access to features, thus
they are only accessible by admin users.

Refs: LIIK-119


mhieta (approved):
All good here :shipit:

PR #131, opened 2020-08-25
Change application root URL to "/city-infra"

mhieta:
We need this in order to move our services under api.hel.fi-domain.
Refs #LIIK-169


mingfeng (approved):
LGTM

PR #132, opened 2020-08-26
Show links to admin change view when clicking features on map view

mingfeng:
The features information are fetched via WMS GetFeatureInfo requests.

For simplicity purposes, we made following assumptions to the format
of WMS GetFeatureInfo responses for all layers (base on current
GeoServer setup):

- All layers are traffic control layers
- The "id" field of features are in the format of "<feature_type>.<instance_id>"
- The <feature_type> must match the model contenttype name after removing all "_"

Refs: LIIK-78

map-view-admin-link


mhieta (changes requested):
Looks good, but could you update the CHANGELOG too.

mhieta (approved):
LGTM

PR #133, opened 2020-08-31
Remove color attribute from traffic signs

ehaivala:
Remove color attribute from TrafficSignPlan and TrafficSignReal models
and their admin interfaces.

Refs LIIK-174


mhieta (approved):
Looks good 👍

PR #134, opened 2020-08-31
Turn map view into a React app with Material UI

mingfeng:
As we are adding more functionality to the map view and there's increasing demand on UI components, it becomes more difficult to develop and maintain the map view with a plain app. Thus we turn the map view into a React App with Material UI components.

Following is a short description of the React app layout:
`api` - singleton services that interact API endpoints
`common` - singleton services that provide common functionality (currently only the map view) to UI components
`components` - React UI components
`models.ts` - interfaces and other type definitions

Refs: LIIK-180

react-map-view


mhieta (approved):
Looks good :shipit:

PR #135, opened 2020-09-01
Update static and media to be configurable

mhieta:
Allow STATIC_URL and MEDIA_URL to be configurable through env-variables


vikoivun (approved):
Ok.

PR #138, opened 2020-09-01
Move map-view build to Dockerfile

mingfeng:
Including the map-view build in entry point slows down the dev server startup quite a bit.

Refs: LIIK-180

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -7,3 +7,5 @@ var/
 *.csv
 /tmp/*
 .*
+**/node_modules
+**/build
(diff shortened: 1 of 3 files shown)


mhieta (changes requested):
This resolves the case with Azure yes, but the overall build process seems to be still quite slow.
I wonder could we optimise our Dockerfile even more, like probably even remove `chown`-commands, since they are now issue due to so many files.

Moreover, we should probably use docker mounts for development-purposes instead of different layers.

mingfeng:
@mhieta Yeah, the map-view build process should probably belong to different folder as we only need the build artifacts for django app.

IMO the development layer is quite useful for development as it can print debug info if there's anything goes wrong. So I'd say let's keep it.

mhieta (approved):
Good

PR #139, opened 2020-09-03
Add source_name and source_id fields to traffic control models

mingfeng:
We need to add null=True and default=None to source_id and
source_names fields as each pair of source_name and source_id
must be unique.

Refs: LIIK-175


mhieta (changes requested):
Could you update CHANGELOG also?

mhieta (approved):
Good :shipit:

PR #140, opened 2020-09-04
Feature: Owner model

ehaivala:
Migrate from all different traffic control device models having text based owner to `Owner` model and a foreign key relation to that instead.

Initialize the database with four default `Owner` instances: city of Helsinki, state, private, and unknown.

Naively migrate from the text representation to one of the default instances. If the text representation doesn't include magic words "helsin", "valtio", or "yksityi" then create new owner instance for the text representation. Remove the old text based owner field from the database after the data migration.

Implement API endpoints, write tests for them, and update all existing functionality and tests to work with the owner relation.

Add admin interface for `Owner` model and add list filter to related model admin views so the list views can be filtered based on ownership.

Refs LIIK-173


mhieta (approved):
Great stuff! :shipit:

PR #141, opened 2020-09-04
Add a management command import operational areas (contract areas) from Helsinki City WFS

mingfeng:
We also updated the OperationalArea model to include more fields from WFS features. The migration assumes that there's no operational areas in database.

Refs: LIIK-166


mhieta (approved):
All good here

mhieta:
Now tests work also. Good job! 👌

PR #142, opened 2020-09-07
Open attached files in new tab

ehaivala:
Implement custom `AdminFileWidget` that adds `target="_blank"` attribute
for the `a` tag. Otherwise the template is identical with the template
used by the default widget.

Use the custom `AdminFileWidget` in file model inlines for any plan and
real models.

Refs LIIK-176


mhieta (approved):
Good

PR #145, opened 2020-09-08
Use raw_id_fields in admin for plan and real relations

ehaivala:
Admin add/change views get really slow when there's a foreign key
relation with lot of choices (20k+ in this case). To remedy this change
admin to use raw_id_fields for any relation to any traffic control real
or plan models since all of those will most likely end up having so many
choices that it will affect the usability considerably.

Refs LIIK-177


mhieta (approved):
Great stuff!

PR #146, opened 2020-09-08
List unreleased changes under correct title in changelog

ehaivala:
Unreleased changes were incorrectly added under v1.1.0 when those should
have been listed as unreleased. Move entries that are not released from
v1.1.0 to unreleased to fix this error.


mhieta (approved):
Great! :shipit:

PR #152, opened 2020-09-29
Add traffic control device types filtering by target model

mhieta:
Update changelog.
Refs #LIIK-191


aslani (approved):
Looks good to me 👍

PR #153, opened 2020-10-13
Add parking area

mingfeng:
Create parking area model based on WFS layer `Pysakointipaikat_alue`

The WFS layer fields are mapped as following:
```
id --> source_id
luokka --> Category.id                    
luokka_nimi --> Category.name         
tyyppi --> area_type
kausi --> season
asukaspysakointitunnus --> resident_parking_id
paikan_asento --> place_position
voimassaolo --> validity
kesto --> duration
pinta_ala --> surface_area
paikat_ala --> parking_slots
lisatieto  --> additional_info
pysayttamiskielto --> stopping_prohibited
paivitetty_tietopalveluun --> updated_at
datanomistaja --> owner
```

Refs: LIIK-193


mhieta (changes requested):
I would change the following naming:

`pysayttamiskielto --> stopping_prohibited`

Also, it seems that `paikat_des` and `paikat_ala` actually means the same thing (but with different specision): how many parking slots are in the area. This value is based on spatial analysis, so it is safer to use lower number here, in this case we should use paikat_ala only and change the naming of the field to `parking_slots`.

mhieta (approved):
Looks good now :shipit:

PR #154, opened 2020-10-14
Add a management command to import parking areas from Helsinki City WFS layer

mingfeng:
This PR adds a management command that can import parking areas from `Pysakointipaikat_alue` WFS layer . The command is idempotent and can be used to update parking areas if there're updates on WFS layer. However, it does not remove parking areas from database if there are parking areas removed from WFS layer.

It should be noted that there are mixed geometries (Polygon & MultiPolygon) in `Pysakointipaikat_alue` WFS layer. We convert all polygons to multi-polygons so that we can have a single geometry type for Parking areas.

This PR also refactor the DataSource mocking that is used both in `test_import_parking_areas_hki_wfs` and `test_import_operational_areas_hki_wfs` to remove code duplication.

Refs: LIIK-193

### DIFF
--- a/traffic_control/management/commands/import_parking_areas_hki_wfs.py
+++ b/traffic_control/management/commands/import_parking_areas_hki_wfs.py
@@ -1,0 +1,86 @@
+from datetime import datetime
+from urllib.request import urlretrieve
+
+from django.contrib.gis.gdal import DataSource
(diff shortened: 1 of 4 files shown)


mhieta (approved):
Good work, no issues here 👌

PR #156, opened 2020-10-23
Update plan model

mingfeng:
- update the help text for `plan_number`
- change `decision_maker` field type from `ForeignKey` to `CharField`. There's a data migration that migrate the decision maker's fullname to the new `CharField`

We also update the requirements files as the build are failing for `django-helusers` because the referenced branch is merged to master.

Note that we also pin a specific version of `djangorestframework` and `ipython` because:
- drf-yasg is not compatible with DRF 3.12+, see details: axnsan12/drf-yasg#641
- ipython 7.17+ requires python 3.7+ (we need to support Python 3.6), see details: https://github.com/ipython/ipython#overview

Refs: LIIK-197


mhieta (approved):
All good here :shipit:

PR #157, opened 2020-10-26
Various improvements to traffic sign importers

mingfeng:
Previously we have extracted additional signs to its own model, but the logic to assign device types were not migrated to new models. This PR adds the same logic to assign the device type of additional sign content instances based on the legacy code in source data.

We also add `source_name` and `source_id` to AbstractAdditionalSignContent  in order to make the imoprter idempotent on additional sign contents.

Refs: LIIK-200


mhieta (approved):
Good

PR #158, opened 2020-10-26
Improve fields naming in Plan API endpoint to make it more understandable

mingfeng:
- rename `plans` to `linked_objects`
- append `_plan_ids` to each object type under `linked_objects`

Refs: LIIK-201


mhieta (approved):
Great job! :shipit:

PR #162, opened 2020-10-28
Improve feature info window

mingfeng:
This PR introduces several improvements to the feature info window including:
- Display txt, code along with id from feature attributes in feature info window
- Use id from attributes to construct the link to open feature admin editing view
- Query all visible layers for feature info

Refs: LIIK-202

Screenshot from 2020-10-28 12-36-23


mhieta (approved):
Good :shipit:

mhieta (changes requested):
Changelog still needs to be updated..

mhieta (approved):
Good

PR #163, opened 2020-10-28
Add localization to map view

mingfeng:
This PR introduced localizations to the map view using i18next and react-i18next. The active language are detected from the url path. And the translations are stored in `map-view/src/locale/` folder.

Refs: LIIK-202

image


mhieta (approved):
All good here

PR #164, opened 2020-11-04
Change planner from a ForeignKey to a CharField

mingfeng:
A data migrations is created to migrate the original planner's display name to the new CharField

Refs: LIIK-211


mhieta (approved):
Good :shipit:

PR #165, opened 2020-11-06
Set Plan linked_objects read only

mingfeng:
Curently there's no easy way to configure different schema for read and write APIs, see details: https://github.com/axnsan12/drf-yasg/issues/70

We mark the linked_objects as read-only to make it more obvious in API docs.

Refs: LIIK-213


mhieta (approved):
Good :shipit:

PR #167, opened 2020-11-09
Improve feature info window to include more information

mingfeng:
Include more description to the feature in the feature info window (see screenshot), and add localization for field labels

The `value` are not displayed if the traffic sign real instance has an empty value for the `value` field

image

Refs: LIIK-208


mhieta (approved):
Good 🚢

PR #172, opened 2020-11-20
Refactor the TrafficSignIconSelect widget

mingfeng:
There wre several issues with previous version of the widget.
The widget hard-coded an id to the icon element, and the id was
referenced in the JavaScript function when updating the icon url.
This caused problems when there are multiple widgets loaded in the
same page. Also the JavaScript code was included in the widget
template, which means it will be included multiple times if there
are multiple widgets in the same page.

This commit includes following changes:
- JavaScript code and css styles are extracted into their own files to make sure they are loaded only once.
- The static files are moved to under traffic_control folder to avoid potential name collisions among apps
- Include icon url as an attribute of select option to simplify the icon url change logic.

Refs: LIIK-217

bugfix-icon-widget


mhieta (approved):
All good here :shipit:

PR #173, opened 2020-11-23
Show an arrow symbol to indicator the direction of traffic signs

mingfeng:
Use up north as the starting direction and rotate the arrow on closewise direction

Refs: LIIK-207

traffic-sign-direction


mhieta (approved):
Nice work! 🚢

PR #174, opened 2020-11-23
Rename ParkingArea to CoverageArea and other related models and fields

mingfeng:
This PR also unify the cases for field and model verbose names to avoid duplicate entries and translation fiels.

Refs: LIIK-218


mhieta (approved):
Good work! :shipit:

PR #178, opened 2020-12-09
 Force 3d geometry when importing from HKI WFS

mingfeng:
We have changed geometry fields to 3d, and source data
from HKI WFS are using 2d geometries. Force 3d coordinates
on those geometries before saving to database.

Refs: LIIK-222


mhieta (approved):
Excellent :shipit:

PR #181, opened 2020-12-14
Add new fields to TrafficControlDeviceType model and update the data csv file

mingfeng:
This PR adds `icon`, `value`, `unit` and `size` fields to `TrafficControlDeviceType` and updated the `traffic_control_deivice_types.csv` data file to include additional fields.

We also update the import command so that it will only update `target_model` field if it has None value on the field or it's an new instance. This is because currently when updating a `TrafficControlDeviceType` instance that has a target model assigned, it always raise an exception (even the target_model field is not updated) and prevents updating the `TrafficControlDeviceType` instance if there are references to the instance. See related code here: https://github.com/City-of-Helsinki/city-infrastructure-platform/blob/8df1b4d2b516da3b92833a5b15b6fc93ac26708d/traffic_control/models/common.py#L306-L333

Refs: LIIK-216


mhieta (approved):
Good

PR #182, opened 2020-12-14
Use value from related device type when saving traffic sign/signpost instances

mingfeng:
This PR adds a mixin class that will use the value from related device type when saving traffic signs and signpost instances. It will only assign the value when the value is empty and there's one available from related device type.

Refs: LIIK-216


mhieta (approved):
All good here, according to latest spec :shipit:

PR #184, opened 2020-12-15
Minimal GHA pipeline for running linters & pytest

vikoivun:
This is still missing Node-side on the tests. But let's have pytest running again.


mhieta (approved):
Appears to be doing its job now as expected and this is sufficient for our usage now.
Great job 👍

PR #185, opened 2020-12-15
Improve AdminTrafficSignIconSelectWidget

mingfeng:
Previously the widget is using the `code` field to construct the icon path for traffic signs, which is not always correct. We now have a new icon field which specifies the right svg file to use for each device type. This PR changes to use the icon field to construct the icon path for the widget.

Since there's case inconsistency between the svg file names and the ones specified in icon field, we changed the svg file names to uppercase and also convert the icon field value into uppercase when using it in the code.

Refs: LIIK-219

traffic-sign-icon


mhieta (approved):
Good

PR #227, opened 2022-03-10
Improve Dockerfile

travenin:
Do installations and possible deletions in a single RUN command. This way that layers don't get too large in size.

However, we could (should) make such multi-stage Dockerfile where the production image is built on top of slim Python. It should contain only build artifacts (and runtime libs), no build-time stuff at all.

Refs LIIK-298


ranta (approved):
LGTM 🚀

PR #236, opened 2022-03-31
Map view improvements

ranta:
- When clicking a Real device, draw it's plan and a line between then
- Show distance from Real to Plan device in meters
- Setting to toggle displaying Plan on map when Real is clicked
- Allow filtering devices by their Project ID

refs. LIIK-266, LIIK-261

image

### DIFF
--- a/map-view/src/common/Map.ts
+++ b/map-view/src/common/Map.ts
@@ -232,6 +232,25 @@ class Map {
     this.extraVectorLayer.getSource()!.clear();
   }
 
+  applyProjectFilters(layerConfig: LayerConfig, projectId: string) {
+    const { sourceUrl } = layerConfig;
+    let wfsFilter = "";
+    if (projectId) {
+      wfsFilter = `&FILTER=%3CFilter%3E%3CPropertyIsLike%20wildCard=%27*%27%20singleChar=%27.%27%20escapeChar=%27!%27%3E%3CPropertyName%3Eproject_id%3C/PropertyName%3E%3CLiteral%3E${projectId}%3C/Literal%3E%3C/PropertyIsLike%3E%3C/Filter%3E`;
+    }
+
+    // Override layer source to apply the filter
+    for (const [identifier, layer] of Object.entries(this.overlayLayers)) {
+      const vectorSource = new VectorSource({
+        format: this.geojsonFormat,
+        url:
+          sourceUrl +
+          `?SERVICE=WFS&VERSION=2.0.0&REQUEST=GetFeature&OUTPUTFORMAT=geojson&TYPENAMES=${identifier}${wfsFilter}`,
+      });
+      layer.setSource(vectorSource);
+    }
+  }
+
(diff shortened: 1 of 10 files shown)


map-view/src/common/Map.ts
```diff
@@ -83,31 +91,124 @@ class Map {
...
+        const featureType: string = feature["id_"].split(".")[0].replace("real", "plan");
+        const feature_layer = overlayConfig["layers"].find((l) => l.identifier === featureType);
+
+        if (feature_layer) {
+          const vectorSource = new VectorSource({
+            format: this.geojsonFormat,
+            url:
+              overlayConfig.sourceUrl +
+              `?SERVICE=WFS&VERSION=2.0.0&REQUEST=GetFeature&OUTPUTFORMAT=geojson&TYPENAMES=${feature_layer.identifier}
+&FILTER=%3CFilter%3E%3CResourceId%20rid=%22${feature_layer.identifier}.${feature.values_.device_plan_id}%22/%3E%3C/Filter%3E`,
```
travenin, line 144:
```ts
              overlayConfig.sourceUrl +
              `?SERVICE=WFS&VERSION=2.0.0&REQUEST=GetFeature&OUTPUTFORMAT=geojson&TYPENAMES=${feature_layer.identifier}` +
              `&FILTER=%3CFilter%3E%3CResourceId%20rid=%22${feature_layer.identifier}.${feature.values_.device_plan_id}%22/%3E%3C/Filter%3E`,
```

map-view/src/common/Map.ts
```diff
@@ -83,31 +91,124 @@ class Map {
...
+          });
+          this.extraVectorLayer.setSource(vectorSource);
+
+          // Add a line between the plan and real points
+          vectorSource.on("featuresloadend", (featureEvent) => {
+            const features = featureEvent.features;
+            if (features) {
+              // @ts-ignore
```
travenin, line 152: Instead of ignoring types we could possibly add some types, maybe

map-view/src/common/Map.ts
```diff
@@ -83,31 +91,124 @@ class Map {
...
+              const real_location = feature.values_.geometry.flatCoordinates;
+              const lineString = new LineString([plan_location, real_location]);
+              const olFeature = new OlFeature({
+                geometry: lineString,
+                name: "Line",
+              });
+              this.extraVectorLayer.getSource()!.addFeature(olFeature);
+
+              // Calculate distance between Real and Plan
+              const distance =
```
travenin, line 164: This could be a separate function with some info about rounding the value to precision of two decimals

map-view/src/common/Map.ts
travenin: Actually I think we could use [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) or something to build queries

travenin (changes requested):

travenin (approved):

PR #239, opened 2022-04-05
Fix deprecation warnings coming from Cityinfra code base

travenin:
We should update hel-users to fix warnings coming from there

```
.../site-packages/helusers/apps.py:8
  /opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/helusers/apps.py:8: RemovedInDjango40Warning: django.utils.translation.ugettext_lazy() is deprecated in favor of django.utils.translation.gettext_lazy().
    verbose_name = _("Helsinki Users")
...
```

### DIFF
--- a/city-infrastructure-platform/urls.py
+++ b/city-infrastructure-platform/urls.py
@@ -1,8 +1,7 @@
 from django.conf import settings
-from django.conf.urls import url
 from django.conf.urls.i18n import i18n_patterns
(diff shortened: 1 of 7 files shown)


ranta (approved):
Try to aim for max 72 characters on commit messages, GitHub wraps messages longer than that to a new line, which looks a bit ugly :D

PR #240, opened 2022-04-11
Add Traffic and Additional signs to Map View

ranta:
- Add Traffic and Additional signs WFS API Endpoints
- Add Traffic and Additional signs to Map View
- Fix WFS filtering when filter string is empty
- Allow layer filter fields to be configured
- Add feature clustering for Map View

image


map-view/src/api/__mocks__/mock-data.ts
```diff
@@ -31,4 +31,5 @@ export const mockMapConfig: MapConfig = {
       },
     ],
   },
+  traffic_sign_icons_url: "http://127.0.0.1:8000/static/traffic_control/svg/traffic_sign_icons/",
```
travenin, line 34: This should be relative to app url root, not full url.

travenin (changes requested):

map-view/src/api/__mocks__/mock-data.ts
travenin: I need to read better :eyes: this is a mock file anyway.

travenin (approved):

PR #243, opened 2022-04-13
Don't apply migrations in production image

travenin:
Add `check-migrations.sh` and `apply-migrations.sh` scripts which hold
required commands. In OpenShift we will have `job`'s which run these
separately before starting API service pod(s). Pods only check that
migrations are applied before startup.


ranta (approved):
LGTM

PR #265, opened 2022-06-13
Hide user ids from object in anonymous requests

travenin:
Hide created_by, updated_by and deleted_by from non-logged in users for data privacy, even when user ids are just UUID values.

Refs LIIK-367


ranta (approved):
LGTM 🚀

PR #267, opened 2022-06-29
Use drf-spectacular for OpenAPI 3 schema generation

travenin:


travenin:
This PR was replaced with #307

PR #419, opened 2024-03-13
Remove last file from city-infrastructure-platform

jforsman:
Refs LIIK-585


harriris-vincit:
I'd suggest to start using conventional commits in this project too: https://www.conventionalcommits.org/en/v1.0.0/#specification. 

So then here the commit message could be "chore: remove last file from city-infrastructure-platform". Otherwise LGTM and can approve once decided if conventional commits are to be taken in use.

PR #631, opened 2025-08-06
feat: Remove order fields from all models that have it

jforsman:
Height fields is used instead, order is just duplicate information so it can be removed.

Refs: LIIK-416


shundread (approved):
lgtm

PR #837, opened 2026-04-21
feat: Move signs that are actually signpost from traffic signs to sig…

jforsman:
…npost tables

* During first import it was decided that some signpost should be place to trafficsign tables
* This has now changed

Refs: LIIK-943


traffic_control/management/commands/move_ticket_machines_to_additional_signs.py
```diff
@@ -272,20 +232,7 @@ def _create_plan_detail_record(
...
-            had_mount_type=ts_plan.mount_type is not None,
-            had_road_name=bool(ts_plan.road_name),
-            had_lane_number=ts_plan.lane_number is not None,
-            had_lane_type=ts_plan.lane_type is not None,
-            had_location_specifier=ts_plan.location_specifier is not None,
-            had_validity_period_start=ts_plan.validity_period_start is not None,
-            had_validity_period_end=ts_plan.validity_period_end is not None,
-            had_source_name=bool(ts_plan.source_name),
-            had_source_id=bool(ts_plan.source_id),
+            **self._get_plan_field_tracking_kwargs(ts_plan),
```
shundread: I would personally put the `**spread_stuff` at the beginning of the `create` statement, as the current pattern could lead to changes in the base class producing unexpected changes in the derived class. Normally when we're overriding stuff we expect the new values to be the ones that stay

```python
defaults = {"present_in_both_dicts": "DEFAULT VALUE"}

{"foo": "bar", "present_in_both_dicts": "OVERRIDEN VALUE", **defaults}
# {'foo': 'bar', 'present_in_both_dicts': 'DEFAULT VALUE'}

{**defaults, "foo": "bar", "present_in_both_dicts": "OVERRIDEN VALUE"}
# {'present_in_both_dicts': 'OVERRIDEN VALUE', 'foo': 'bar'}
```

traffic_control/management/commands/move_ticket_machines_to_additional_signs.py
```diff
@@ -325,16 +272,7 @@ def _create_real_detail_record(
...
-            had_size=ts_real.size is not None,
-            had_direction=ts_real.direction is not None,
-            had_reflection_class=ts_real.reflection_class is not None,
-            had_surface_class=ts_real.surface_class is not None,
-            had_mount_type=ts_real.mount_type is not None,
-            had_road_name=bool(ts_real.road_name),
-            had_lane_number=ts_real.lane_number is not None,
-            had_lane_type=ts_real.lane_type is not None,
-            had_location_specifier=ts_real.location_specifier is not None,
+            **self._get_real_field_tracking_kwargs(ts_real),
```
shundread: Same complaint about `**spread_stuff` positioning in an overriden method.

traffic_control/management/commands/move_traffic_signs_to_signposts.py
```diff
@@ -0,0 +1,354 @@
...
+        """Output any additional warnings specific to the migration."""
+        pass  # No additional warnings for signpost migration
+
+    def get_lost_data_report_filename(self) -> str:
+        """Get filename for lost data report."""
+        return f"signpost_migration_lost_fields_{datetime.now().strftime('%Y-%m-%d')}.csv"
+
+    # Migration-specific methods
+
+    # Migration-specific methods
```
shundread: Duplicate comment of comment 2 lines above

traffic_control/management/commands/move_traffic_signs_to_signposts.py
```diff
@@ -0,0 +1,354 @@
...
+        SignpostMigrationPlanRecord.objects.create(
+            migration_run=self.migration_run,
+            original_traffic_sign_plan=ts_plan,
+            new_signpost_plan=new_plan,
+            original_id=ts_plan.id,
+            new_id=new_plan.id if new_plan else None,
+            device_type_code=ts_plan.device_type.code,
+            had_mount_plan=ts_plan.mount_plan is not None,
+            had_plan=ts_plan.plan is not None,
+            **self._get_plan_field_tracking_kwargs(ts_plan),
```
shundread: Spread operation positioning

traffic_control/management/commands/move_traffic_signs_to_signposts.py
```diff
@@ -0,0 +1,354 @@
...
+            migration_run=self.migration_run,
+            original_traffic_sign_real=ts_real,
+            new_signpost_real=new_real,
+            original_id=ts_real.id,
+            new_id=new_real.id if new_real else None,
+            device_type_code=ts_real.device_type.code,
+            plan_mapping_found=plan_mapping_found,
+            had_mount_real=ts_real.mount_real is not None,
+            had_traffic_sign_plan=ts_real.traffic_sign_plan is not None,
+            **self._get_real_field_tracking_kwargs(ts_real),
```
shundread: Spread operation positioning

traffic_control/management/commands/move_traffic_signs_to_signposts.py
```diff
@@ -0,0 +1,354 @@
...
+                    updated_at=ts_plan.updated_at,
+                    created_by=ts_plan.created_by,
+                    updated_by=ts_plan.updated_by,
+                )
+                self.plan_id_mapping[str(ts_plan.id)] = str(new_plan.id)
+                files = TrafficSignPlanFile.objects.filter(traffic_sign_plan=ts_plan)
+                file_count = self._migrate_files(files, SignpostPlanFile, "signpost_plan", new_plan)
+                self.stats["plan_files_migrated"] += file_count
+
+                # Create tracking record BEFORE deletion (so original object still exists)
```
shundread: This is not a request for change, but a request for clarification.

I'm curious about the distinct `_create_plan_detail_record` logic in here vs `move_ticket_machines_to_additional_signs.py` command.

```python
# move_traffic_signs_to_signpost.py
            if not dry_run:
                (...)

                # Create tracking record BEFORE deletion (so original object still exists)
                self._create_plan_detail_record(ts_plan, new_plan, file_count)

                # Now safe to delete
                self._delete_original_traffic_sign_real(ts_plan, hard_delete, system_user)
                self.stats["plans_migrated"] += 1
                self.stdout.write(f"    ✓ Created SignpostPlan {new_plan.id}")
            else:
                # Dry-run: still create detail record for tracking
                self._create_plan_detail_record(ts_plan, new_plan, file_count)


# move_ticket_machines_to_additional_signs
            if not dry_run:
                (...)

                # Delete original based on mode
                self._delete_original_traffic_sign_real(ts_plan, hard_delete, system_user)

                self.stats["plans_migrated"] += 1
                self.stdout.write(f"    ✓ Created AdditionalSignPlan {new_plan.id}")

            # Create detail record (outside dry_run check so we log even in dry-run)
            self._create_plan_detail_record(ts_plan, new_plan, parent_sign, multiple_parents, file_count)
```

Why does the deletion of the original matter here and why does it not matter in the other command?

traffic_control/management/commands/move_traffic_signs_to_signposts.py
```diff
@@ -0,0 +1,354 @@
...
+
+        new_real = None
+        file_count = 0
+        if not dry_run:
+            new_real, file_count = self._create_signpost_real(ts_real, signpost_plan_id, hard_delete, system_user)
+            # Create tracking record BEFORE deletion (so original object still exists)
+            self._create_real_detail_record(ts_real, new_real, plan_mapping_found, file_count)
+
+            # Now safe to delete
+            self._delete_original_traffic_sign_real(ts_real, hard_delete, system_user)
```
shundread, line 287: There's also a mismatch between usage of `_delete_original_traffic_sign_real` in both commands. Here it's left to the responsibility of `_process_single_traffic_sign_real`, which seems like a more logical place to put it, while in the other command it's left to `_create_additional_sign_real`, which sounds like a strange side-effect of creating something.

shundread (changes requested):
Partial review for the time being just so you're not left waiting and since you're still submitting changes yourself. I can do a complete review later.

traffic_control/management/commands/move_traffic_signs_to_signposts.py
jforsman: Done

traffic_control/management/commands/move_ticket_machines_to_additional_signs.py
jforsman: sure, can do that

traffic_control/management/commands/move_traffic_signs_to_signposts.py
jforsman: It should be the same in both, will check if there was some hidden detailed stuff with these

jforsman: Logic changed in both, and now they should be the same.

traffic_control/admin/common.py
```diff
@@ -19,6 +27,178 @@
...
+        """
+        if obj.success and obj.dry_run:
+            bg, text = "#17a2b8", "✓ DRY RUN SUCCESS"
+        elif obj.success:
+            bg, text = "#28a745", "✓ SUCCESS"
+        elif obj.error_message:
+            bg, text = "#dc3545", "✗ FAILED"
+        else:
+            bg, text = "#6c757d", "⋯ IN PROGRESS"
+        return mark_safe(
```
shundread: `mark_safe(f"words {foo} more words {bar}")` -> `format_html("words {} more words {}", foo, bar)` or `format_html("words {foo} more words {bar}", foo=val1, bar=val2)`

traffic_control/admin/signpost_migration.py
```diff
@@ -0,0 +1,665 @@
...
+                'border-radius: 3px; font-weight: bold;">HARD DELETE</span>'
+            )
+        else:
+            return mark_safe(
+                '<span style="background-color: #28a745; color: #fff; padding: 3px 8px; '
+                'border-radius: 3px; font-weight: bold;">SOFT DELETE</span>'
+            )
+
+    @admin.display(description=_("Duration"))
+    def duration(self, obj: SignpostMigrationRun) -> str:
```
shundread: The implementation for this display field is nearly identical here an in `TicketMachineTrafficSignMigrationRunAdmin`. It's a difference of `obj.started_at` vs `obj.executed_at` and type hints. Is this a hint that we should be using the same field name to indicate the start of the migration and then share the implementation for this field in `TrafficSignMigrationRunAdminMixin`?

traffic_control/admin/signpost_migration.py
```diff
@@ -0,0 +1,665 @@
...
+            {
+                "fields": ("error_message",),
+                "classes": ("collapse",),
+            },
+        ),
+    )
+    inlines = [SignpostMigrationPlanRecordInline, SignpostMigrationRealRecordInline]
+
+    @admin.display(description=_("Mode"))
+    def mode_display(self, obj: SignpostMigrationRun) -> str:
```
shundread: Identical implementation in `TicketMachineTrafficSignMigrationRunAdmin`, should it move to `TrafficSignMigrationRunAdminMixin`?

shundread (changes requested):
A couple more small change requests.

shundread (approved):

shundread (changes requested):
Can you check if wrapping the transaction with a `with set_actor(get_system_user()):` block improves the logging of who's creating new objects in the auditlog?

jforsman:
Add the with set_actor(system_user) to handle function of the base management command, and yes it sets the actor for the LogEntry, so lets keep it like that.

shundread (approved):
Good to go

PR #840, opened 2026-04-25
Allow editing plan replacements in admin via service layer

Romadeoliveira3:
This PR addresses the following TODO:

Modifying replacements can be allowed when Admin UI uses service layer function

To solve this, I updated the Admin UI flow so plan replacement changes are handled through the service layer instead of direct model writes. This keeps the admin behavior aligned with the business rules already enforced in the service layer and API.

What was done:

- Added support for editing `replaces` directly from the admin form.
- Routed replacement updates through the plan update service.
- Preserved service-layer validation for all replacement rules.
- Added automated test coverage for both success and error scenarios.
- Preserved the original TrafficSign-focused tests and added new tests for broader replaceable plan coverage.

This change allows replacement updates from the Admin UI while keeping the logic centralized and consistent.

About me:
My name is Romario Jonas, but you can call me Roma. I am a Brazilian Full Stack developer contributing to Finnish open-source projects to strengthen my technical skills in an international environment, improve my English, and learn more about the local engineering culture. My goal is to learn through real collaboration and contribute in a respectful way.

Thank you for your time and feedback.

### DIFF
--- a/traffic_control/services/common.py
+++ b/traffic_control/services/common.py
@@ -1,5 +1,5 @@
 from datetime import timedelta
-from typing import Callable, Type
+from typing import Callable, Optional, Type
 from uuid import UUID
 
 from django.core.exceptions import ValidationError
@@ -7,6 +7,7 @@ from django.db import transaction
 from django.db.models import Model, Q
 from django.utils import timezone
 
+from django.utils.translation import gettext_lazy as _
 from traffic_control.enums import Lifecycle
 from traffic_control.mixins.models import SoftDeleteModel
 from users.models import User
@@ -103,6 +104,28 @@ def device_plan_update(
     return instance
 
 
+def validate_device_plan_replacement(old: SoftDeleteModel, new: Optional[SoftDeleteModel]):
+    """
+    Validate that a device plan can be replaced by another device plan.
+
+    :param old: The old device plan to be replaced.
+    :param new: The new device plan that will replace the old one.
+    :raises ValidationError: If the replacement cannot be performed.
+    """
+    if old.replaced_by and (not new or not new.pk or old.replaced_by.pk != new.pk):
+        raise ValidationError(_("Cannot replace a device plan that is already replaced"))
+
+    if new and new.pk:
+        if old.pk == new.pk:
+            raise ValidationError(_("Cannot replace a device plan with itself"))
+
+        check_replaced = old.replaces
+        while check_replaced:
+            if check_replaced.pk == new.pk:
+                raise ValidationError(_("Cannot form a circular replacement chain"))
+            check_replaced = check_replaced.replaces
+
+
 @transaction.atomic
 def device_plan_replace(
     *,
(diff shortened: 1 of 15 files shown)


traffic_control/services/common.py
```diff
@@ -103,6 +104,28 @@ def device_plan_update(
...
 
+def validate_device_plan_replacement(old: SoftDeleteModel, new: Optional[SoftDeleteModel]):
+    """
+    Validate that a device plan can be replaced by another device plan.
+
+    :param old: The old device plan to be replaced.
+    :param new: The new device plan that will replace the old one.
+    :raises ValidationError: If the replacement cannot be performed.
+    """
+    if old.replaced_by and (not new or not new.pk or old.replaced_by.pk != new.pk):
```
shundread, line 115: We AI-generate a lot of code in this project, but it looks like this change wasn't reviewed.

The worst offense here is that this check offers no benefit over the `if old.replaced_by` check other than apparently trying to let the "replace with itself" case fall through to be caught by the next check.

Validation should either stop at the first error (best performance) or in cases where richer information is often needed offer a full report of everything that is wrong in the validation. This does neither.

In addition, someone familiar with the system would immediately spot that none of the relevant traffic control or city furniture objects will ever not have a pk because we're using UUIDs and these will necessarily get created as part of the object's instantiation sequence even if it hasn't been written to the database yet. So there's no need to ever check if `obj.pk` exists for these objects.

traffic_control/tests/factories.py
```diff
@@ -419,10 +420,12 @@ def get_road_marking_real(
     traffic_sign_real=None,
 ) -> RoadMarkingReal:
     user = get_user("test_user")
+    road_marking_plan = road_marking_plan or get_road_marking_plan()
```
shundread, line 423: This is also typical of AI code generation, not a problem by itself but increases diff size and line code count by doing a pointless replacement in the form of

```diff
-f(value=a or b)
+value = a or b
+f(value=value)
```

shundread (commented):
Greetings Roma, I appreciate the enthusiasm but we have an internal process for prioritizing tasks and we don't currently accept contributions from people outside the organization.

I skimmed through it and added a couple of comments on the more glaring stuff that would make me request further changes right away, but I didn't really look deeper into the entire pull request, so this review is going to be incomplete, sorry.

We will close the pull request without merging in accordance to the project's current policy.

If you're interested in participating on open source projects with the city of Helsinki I suggest you search for information whether the city itself is hiring developers or finding out which companies the city is partnering with in order to develop its services and applying to those.

shundread:
Reasons for closing the pull request described [here](https://github.com/City-of-Helsinki/city-infrastructure-platform/pull/840#pullrequestreview-4203102407).

Romadeoliveira3:
Hi Shundread,

Thank you for the detailed feedback and for taking the time to review my PR, even with the current internal policy.

I completely understand the decision to close it. This was a great learning exercise for me to understand the Finnish engineering culture and the City's infrastructure code and technical standards.

Regarding your technical points:

UUID and PK: That’s a great catch. I wasn't aware that the UUID instantiation happened so early in the object sequence. I should have studied the structure more deeply; sorry for that.

I appreciate the advice about the hiring process and partner companies. I will definitely look into that, as I really admire the work being done in Helsinki.

Thanks again for the guidance!

Best regards,

PR #841, opened 2026-04-27
2025 streetscan analyze

jforsman:


.github/copilot-instructions.md
```diff
@@ -9,7 +9,7 @@
 ## Code Style & Quality
 - **Type Hinting**: Always use explicit Type Hinting for all Python function signatures (parameters and return types).
 - **Complexity**: Maintain a maximum cognitive complexity of 10.
-- **Return Statements**: Use a maximum of **4 return statements** per function. If more are needed, refactor the logic.
+- **Return Statements**: Use a maximum of **3 return statements** per function. If more are needed, refactor the logic into helper methods.
```
shundread, line 12: Doesn't belong in this PR but you can leave it there.

traffic_control/analyze_utils/traffic_sign_data.py
```diff
@@ -729,14 +729,14 @@ def _check_device_type(self, row, object_type="additionalsign") -> Union[ImportR
                 reason=f"device type code not found: {row[CSVHeaders.code]}",
             )
 
-    @staticmethod
-    def _check_ticket_machine(row) -> Union[ImportResult, None]:
+    def _check_ticket_machine(self, row) -> bool:
```
shundread: I don't think there is a need to remove `@staticmethod` here

traffic_control/analyze_utils/traffic_sign_data_v2_code_transform.py
```diff
@@ -0,0 +1,457 @@
...
+
+        Args:
+            row (dict): CSV row dictionary (modified in place).
+        """
+        code = row.get(CSVHeadersV2.code, "").strip()
+        if code in INTERNAL_ADDITIONAL_INFO_ENRICHMENTS:
+            row["internal_additional_info"] = INTERNAL_ADDITIONAL_INFO_ENRICHMENTS[code]
+
+    @staticmethod
+    def _is_skipable_code(code: str) -> bool:
```
shundread: ~skipable~ skippable

traffic_control/analyze_utils/traffic_sign_data_v2_code_transform.py
```diff
@@ -0,0 +1,457 @@
...
+    LOCATION_SPECIFIER_4_CODES,
+    NUMBER_CODE_DEPENDENT_CODES,
+    NUMBER_CODE_PATTERN,
+)
+
+
+class CodeTransformMixin:
+    """Mixin providing sign code filtering and enrichment pipeline methods."""
+
+    def _should_be_filtered(self, row: dict, filter_f: Callable, reason: str) -> bool:
```
shundread: This method, along with `_filter_and_enrich_sign_rows` are really confusing to read.

The name `_should_be_filtered` reads like  all the method would do is give me a True / False result after looking what's in the rows, but the the method alters the object's `filtered_signs` field as a side-effect.

On `_filter_and_enrich_sign_rows`, there's a local `filtered_rows` variable, but that value seems to mean `"enriched_rows"` instead, because that's how it's used in `TrafficSignAnalyzerV2`'s `__init__` method.

I don't know what's the "good" logic flow for this, but this definitely isn't it.

traffic_control/analyze_utils/traffic_sign_data_v2_code_transform.py
```diff
@@ -0,0 +1,457 @@
...
+                    "source_id": row.get(CSVHeadersV2.id),
+                    "code": code,
+                    "reason": reason,
+                    "csv_row": self._row_to_csv_line(row, self.delimiter),
+                }
+            )
+            return True
+        return False
+
+    def _apply_direct_code_replacement(self, row: dict) -> str:
```
shundread: Why the return value if it edits the object in-place and the return value is never used?

This is a problem in many methods in this class.

shundread (changes requested):
This is messy. Very partial review, but at least a few things could be addressed before I take a better look.

traffic_control/analyze_utils/traffic_sign_data_v2_code_transform.py
jforsman: Yeah, probably won't change to useing immutables as the memory consumption would go throught the roof, the non use return values probably gonna change.

traffic_control/analyze_utils/traffic_sign_data_v2_code_transform.py
jforsman: Yes... changed the name to processed_rows. Add mention where the transformations are documented in. 
If you don't have any better idea on how to do this then I won't touch it more :).

jforsman: should_be_filtered also removed totally, replaced with _record_filtered_sign method.

traffic_control/analyze_utils/traffic_sign_data_v2_db_builders.py
```diff
@@ -0,0 +1,301 @@
...
+            csv_file_path (str): Path to the CSV file.
+            delimiter (str): CSV delimiter character to use.
+
+        Returns:
+            list[dict]: List of dictionaries representing CSV rows.
+        """
+        rows = []
+        with open(csv_file_path) as f:
+            reader = csv.DictReader(f, delimiter=delimiter)
+            for row in reader:
```
shundread: Would it be better to have the implementation of this function be something like:

```python
with open(csv_file_path):
    reader = csv.DictReader(f, delimiter=delimiter)
    return [{k: row[k].strip() for k in row} for row in reader]
```

So we run these '.strip()` calls in only one place and don't have to keep doing that everywhere else?

shundread: For further reference:

```
$ git diff master | grep strip
                     filter(lambda x: not x[CSVHeaders.parent_sign_id].strip(), self.additional_signs_by_id.values()),
         if not row[CSVHeaders.parent_sign_id].strip():
+        code = row.get(CSVHeadersV2.code, "").strip()
+        code = row.get(CSVHeadersV2.code, "").strip()
+        color_value = row.get(CSVHeadersV2.color, "").strip()
+        code = row.get(CSVHeadersV2.code, "").strip()
+        color_value = row.get(CSVHeadersV2.color, "").strip()
+        code = row.get(CSVHeadersV2.code, "").strip()
+        number_code_value = row.get(CSVHeadersV2.number_code, "").strip()
+        code = row.get(CSVHeadersV2.code, "").strip()
+        number_code_value = row.get(CSVHeadersV2.number_code, "").strip()
+        code = row.get(CSVHeadersV2.code, "").strip()
+        location_specifier_value = row.get(CSVHeadersV2.location_specifier, "").strip()
+        code = row.get(CSVHeadersV2.code, "").strip()
+            code = row.get(CSVHeadersV2.code, "").strip()
+            code = row.get(CSVHeadersV2.code, "").strip()
+            source_id = row.get(CSVHeadersV2.id, "").strip()
+            code = row.get(CSVHeadersV2.code, "").strip()
+                lambda x: x.get(CSVHeadersV2.mount_id).strip() not in self.mounts_by_id.keys(),
+            status = obj.get(CSVHeadersV2.status, "").strip().lower()
+        return {row.get(CSVHeadersV2.id, "").strip() for row in csv_rows if row.get(CSVHeadersV2.id, "").strip()}
+        return output.getvalue().strip()
+        if not mount_id.strip():
+        for obj in filter(lambda x: not x[CSVHeadersV2.parent_sign_id].strip(), self.additional_signs_by_id.values()):
+            parent_source_id = sign_data.get(CSVHeadersV2.parent_sign_id, "").strip()
+            mount_source_id = sign_data.get(CSVHeadersV2.mount_id, "").strip()
+            mount_source_id = sign_data.get(CSVHeadersV2.mount_id, "").strip()
+                if sign.get(CSVHeadersV2.mount_id, "").strip() == mount_source_id
+                if sign.get(CSVHeadersV2.mount_id, "").strip() == mount_source_id
+            code = sign_data.get(CSVHeadersV2.code, "").strip()
+            code = sign_data.get(CSVHeadersV2.code, "").strip()
+            status = sign_data.get(CSVHeadersV2.status, "").strip()
+            internal_status = sign_data.get("internal_status", "").strip()
+                    "csv_code": sign_data.get(CSVHeadersV2.code, "").strip(),
+                    "csv_ssurl": sign_data.get(CSVHeadersV2.attachment_url, "").strip(),
+                    sign.get(CSVHeadersV2.code, "").strip()
+                    if sign.get(CSVHeadersV2.code, "").strip()
+                    sign.get(CSVHeadersV2.code, "").strip()
+                    if sign.get(CSVHeadersV2.code, "").strip()
```

traffic_control/analyze_utils/traffic_sign_data_v2_code_transform.py
```diff
@@ -0,0 +1,447 @@
...
+            row (dict): CSV row dictionary (modified in place).
+            code (str): Original device type code.
+            expected_number (str): Expected number value to validate against.
+            replacement_code (str): Replacement code if validation passes.
+        """
+        code_parts = code.split("_")
+        if len(code_parts) <= 1:
+            return
+        extracted_number = code_parts[-1]
+        if extracted_number != expected_number:
```
shundread, line 274: Just double-checking, there is no need to record a failure here like it's done in `_apply_number_code_replacement`, right?

traffic_control/analyze_utils/traffic_sign_data_v2_code_transform.py
```diff
@@ -0,0 +1,447 @@
...
+            row (dict): CSV row dictionary (modified in place).
+        """
+        code = row.get(CSVHeadersV2.code, "").strip()
+        if code not in CONDITIONAL_NUMBER_CODE_REPLACEMENTS:
+            return
+        number_code_value = row.get(CSVHeadersV2.number_code, "").strip()
+        match = NUMBER_CODE_PATTERN.match(number_code_value) if number_code_value else None
+        cleaned_number = match.group(1) if match else ""
+        replacements = CONDITIONAL_NUMBER_CODE_REPLACEMENTS[code]
+        if cleaned_number not in replacements:
```
shundread, line 358: Also I suppose no need to record error here either?

traffic_control/analyze_utils/traffic_sign_data_v2_code_transform.py
```diff
@@ -0,0 +1,447 @@
...
+                "field": "internal_additional_info",
+                "old_value": None,
+                "new_value": value,
+            }
+        )
+        row["internal_additional_info"] = value
+
+    @staticmethod
+    def _is_skippable_code(code: str) -> bool:
+        return code in ("6", "7")
```
shundread: This stands out a little bit from other bits of the file in that it has set two of values hardcoded here rather in a constant, even though there are even equally small sets (`ZEBRA_CROSSING_RIGHT_CODES`, two elements) or even smaller sets (`CONDITIONAL_NUMBER_CODE_REPLACEMENTS` one element) defined in the constants file.

No need to fix though, just pointing it out.

traffic_control/analyze_utils/traffic_sign_data_v2_db_builders.py
```diff
@@ -0,0 +1,301 @@
...
+        return {
+            obj.source_id: getattr(obj.device_type, attr)
+            for obj in model_class.objects.filter(
+                source_name__startswith=TrafficSignImporter.SOURCE_NAME,
+                is_active=True,
+            ).select_related("device_type")
+        }
+
+    @staticmethod
+    def _build_sign_reals_by_source_id() -> dict[str, str | None]:
```
shundread, line 145: unused

traffic_control/analyze_utils/traffic_sign_data_v2_db_builders.py
```diff
@@ -0,0 +1,301 @@
...
+        """Build mapping from source_id to device_type.code for active TrafficSignReal objects.
+
+        Returns:
+            dict[str, str | None]: Dictionary mapping source_id to device_type code.
+        """
+        return DbBuilderMixin._build_device_type_attr_map(TrafficSignReal, "code")
+
+    @staticmethod
+    def _build_additional_sign_reals_by_source_id() -> dict[str, str | None]:
```
shundread, line 154: unused

traffic_control/analyze_utils/traffic_sign_data_v2_db_builders.py
```diff
@@ -0,0 +1,301 @@
...
+        """Build mapping from source_id to device_type.code for active AdditionalSignReal objects.
+
+        Returns:
+            dict[str, str | None]: Dictionary mapping source_id to device_type code.
+        """
+        return DbBuilderMixin._build_device_type_attr_map(AdditionalSignReal, "code")
+
+    @staticmethod
+    def _build_sign_reals_legacy_codes_by_source_id() -> dict[str, str | None]:
```
shundread, line 172: unused

traffic_control/analyze_utils/traffic_sign_data_v2_db_builders.py
```diff
@@ -0,0 +1,301 @@
...
+        """Build mapping from source_id to device_type.legacy_code for TrafficSignReal objects.
+
+        Returns:
+            dict[str, str | None]: Dictionary mapping source_id to device_type legacy_code.
+        """
+        return DbBuilderMixin._build_device_type_attr_map(TrafficSignReal, "legacy_code")
+
+    @staticmethod
+    def _build_additional_sign_reals_legacy_codes_by_source_id() -> dict[str, str | None]:
```
shundread, line 181: unused

traffic_control/analyze_utils/traffic_sign_data_v2_db_builders.py
```diff
@@ -0,0 +1,301 @@
...
+        Returns:
+            dict[str, str]: Dictionary mapping source_id to db_id.
+        """
+        return {
+            obj.source_id: obj.id
+            for obj in model_class.objects.filter(source_name__startswith=TrafficSignImporter.SOURCE_NAME)
+        }
+
+    @staticmethod
+    def _build_mount_source_id_to_db_id() -> dict[str, str]:
```
shundread, line 214: unused

traffic_control/analyze_utils/traffic_sign_data_v2_db_builders.py
```diff
@@ -0,0 +1,301 @@
...
+        """Build mapping from source_id to db_id for MountReal objects.
+
+        Returns:
+            dict[str, str]: Dictionary mapping source_id to db_id.
+        """
+        return DbBuilderMixin._build_source_id_to_db_id(MountReal)
+
+    @staticmethod
+    def _build_sign_source_id_to_db_id() -> dict[str, str]:
```
shundread, line 223: unused

traffic_control/analyze_utils/traffic_sign_data_v2_db_builders.py
```diff
@@ -0,0 +1,301 @@
...
+        """Build mapping from source_id to db_id for TrafficSignReal objects.
+
+        Returns:
+            dict[str, str]: Dictionary mapping source_id to db_id.
+        """
+        return DbBuilderMixin._build_source_id_to_db_id(TrafficSignReal)
+
+    @staticmethod
+    def _build_additional_sign_source_id_to_db_id() -> dict[str, str]:
```
shundread, line 232: unused

traffic_control/analyze_utils/traffic_sign_data_v2_db_builders.py
```diff
@@ -0,0 +1,301 @@
...
+
+        Returns:
+            dict[str, str | None]: Dictionary mapping source_id to the device_type attribute value.
+        """
+        return {
+            obj.source_id: getattr(obj.device_type, attr)
+            for obj in model_class.objects.filter(
+                source_name__startswith=TrafficSignImporter.SOURCE_NAME,
+                is_active=True,
+            ).select_related("device_type")
```
shundread, line 141: I did a quick skim of the code and I didn't really find a reason for the `select_related` statement. Does it actually matter?

traffic_control/analyze_utils/traffic_sign_data_v2_data_loading.py
```diff
@@ -0,0 +1,267 @@
...
+        mount_data = self.mounts_by_id.get(data.get(CSVHeadersV2.mount_id), None)
+        if mount_data:
+            mount_point = Point(float(mount_data[CSVHeadersV2.coord_x]), float(mount_data[CSVHeadersV2.coord_y]), 0.0)
+            data["distance_to_mount"] = mount_point.distance(
+                Point(float(data[CSVHeadersV2.coord_x]), float(data[CSVHeadersV2.coord_y]), 0.0)
+            )
+        else:
+            data["distance_to_mount"] = None
+
+    def _get_signs_by_id(self, csv_rows: list[dict], id_field_name: str = CSVHeadersV2.id, flat: bool = False) -> dict:
```
shundread: Every usage of `_get_signs_by_id` uses `flat=True`. Having the parameter just increases cognitive load of reading the code.

traffic_control/analyze_utils/traffic_sign_data_v2_data_loading.py
```diff
@@ -0,0 +1,267 @@
...
+
+        Returns:
+            dict: Dictionary mapping mount IDs to lists of sign data dicts.
+        """
+        grouped: dict = {}
+        for v in signs_dict.values():
+            grouped.setdefault(v[CSVHeadersV2.mount_id], []).append(v)
+        return grouped
+
+    def _get_additional_signs_by_id(
```
shundread: Every usage of `_get_additional_signs_by_id` uses `flat=True`, so we don't need the flag.

traffic_control/analyze_utils/traffic_sign_data_v2_data_loading.py
```diff
@@ -0,0 +1,267 @@
...
+                    data["distance_to_parent"] = None
+                    data["parent_is_additional_sign"] = None
+                    data["parent_code"] = None
+        return additional_signs
+
+    def _get_additional_signs_by_mount_id(self):
+        """Group additional signs by mount ID."""
+        return self._group_by_mount_id(self.additional_signs_by_id)
+
+    def _get_signposts_by_id(
```
shundread: Every usage of `_get_signposts_by_id` uses `flat=True` we don't need that argument.

traffic_control/analyze_utils/traffic_sign_data_v2_data_loading.py
```diff
@@ -0,0 +1,267 @@
...
+        Returns:
+            Point | None: A Point object, or None if coordinates are invalid.
+        """
+        try:
+            return Point(float(csv_row[CSVHeadersV2.coord_x]), float(csv_row[CSVHeadersV2.coord_y]), 0.0)
+        except (KeyError, TypeError, ValueError):
+            return None
+
+    @staticmethod
+    def _get_objects_by_id(
```
shundread: Every direct usage of `_get_objects_by_id`  uses `flat=True` and every indirect usage of `_get_objects_by_id` also calls this with `flat=True`, the method can be simplified.

traffic_control/analyze_utils/traffic_sign_data_v2_data_loading.py
```diff
@@ -0,0 +1,267 @@
...
+            id_field_name (str): Name of the ID field in CSV.
+            flat (bool): If True, store single object per ID.
+
+        Returns:
+            dict: Dictionary of additional signs indexed by ID with calculated distances.
+        """
+        additional_signs = self._get_objects_by_id(
+            csv_rows, id_field_name=id_field_name, filter_f=self._is_additional_sign, flat=flat
+        )
+        for k, data in additional_signs.items():
```
shundread: No idea why SonarCube nor ruff accuse the unused variable.

```python
for data in additional_signs.values():
```

traffic_control/analyze_utils/traffic_sign_data_v2.py
```diff
@@ -0,0 +1,162 @@
...
+        self.signs_by_id = self._get_signs_by_id(self.sign_rows, CSVHeadersV2.id, flat=True)
+        self.signs_by_mount_id = self._get_signs_by_mount_id()
+        self.additional_signs_by_id = self._get_additional_signs_by_id(self.sign_rows, CSVHeadersV2.id, flat=True)
+        self.additional_signs_by_mount_id = self._get_additional_signs_by_mount_id()
+        self.signposts_by_id = self._get_signposts_by_id(self.sign_rows, CSVHeadersV2.id, flat=True)
+        self.signposts_by_mount_id = self._get_signposts_by_mount_id()
+        self._combine_mounts_with_signs(self.mounts_by_id, self.additional_signs_by_mount_id, self.signs_by_mount_id)
+        self.no_mounts_per_sign_id = self._get_non_existing_mounts_by_sign_id()
+        self.no_mounts_per_additional_sign_id = self._get_non_existing_mounts_by_additional_sign_id()
+        self.no_mounts_per_signpost_id = self._get_non_existing_mounts_for(self.signposts_by_id)
```
shundread: The two lines above this one use wrapper functions that just wrap and parametrize a call to `_get_non_existing_mounts_for`, but here we use it directly.

To be honest I this pattern

```python
def specialized_method_a():
    some_method(a_rows)

def specialized_method_b():
    some_method(b_rows)

def some_method(rows):
    do_stuff(rows)

def some_process():
    specialized_method_a()
    specialized_method_b()
```

Is worse than

```python
def some_method(rows):
    do_stuff(rows)

def some_process():
    some_method(a_rows)
    some_method(b_rows)
```

Because while reading `some_process` I cannot immediately assume that `specialized_method_a` and `specialized_method_b` have basically the same implementation. Reviewing this sort of pattern simply forces me to bounce back and forth into the source just to find out they're so similar.

I would prefer to see these one-liner specializations gone unless a method takes a lot of parameters which are easy to get wrong. It would cut down a lot on diff size and make reading the process far more straightforward.

For comparison, `ReportsMixin._build_mountless_report` is somewhat more excusable because the `objects: dict, report_type: str, id_key: str` parameter set is somewhat easy to get wrong.

traffic_control/analyze_utils/traffic_sign_data_v2_data_loading.py
```diff
@@ -0,0 +1,267 @@
...
+
+    def _calculate_distance_to_mount(self, data: dict) -> None:
+        """Calculate distance from an object to its mount and store in data['distance_to_mount'].
+
+        Args:
+            data (dict): CSV row dict for a sign or additional sign (modified in place).
+        """
+        mount_data = self.mounts_by_id.get(data.get(CSVHeadersV2.mount_id), None)
+        if mount_data:
+            mount_point = Point(float(mount_data[CSVHeadersV2.coord_x]), float(mount_data[CSVHeadersV2.coord_y]), 0.0)
```
shundread: Use `_point_from_csv_row`?

traffic_control/analyze_utils/traffic_sign_data_v2_data_loading.py
```diff
@@ -0,0 +1,267 @@
...
+            data["distance_to_mount"] = mount_point.distance(
+                Point(float(data[CSVHeadersV2.coord_x]), float(data[CSVHeadersV2.coord_y]), 0.0)
```
shundread: Use `_point_from_csv_row`?

traffic_control/analyze_utils/traffic_sign_data_v2_data_loading.py
```diff
@@ -0,0 +1,267 @@
...
+            dict: Dictionary of additional signs indexed by ID with calculated distances.
+        """
+        additional_signs = self._get_objects_by_id(
+            csv_rows, id_field_name=id_field_name, filter_f=self._is_additional_sign, flat=flat
+        )
+        for k, data in additional_signs.items():
+            self._calculate_distance_to_mount(data)
+            parent_data = self.signs_by_id.get(data.get(CSVHeadersV2.parent_sign_id), None)
+            if parent_data:
+                parent_point = Point(
```
shundread: Use `_point_from_csv_row`?

traffic_control/analyze_utils/traffic_sign_data_v2_data_loading.py
```diff
@@ -0,0 +1,267 @@
...
+from django.contrib.gis.geos import Point
+
+from .traffic_sign_data_v2_constants import CSVHeadersV2, VALID_STATUS_VALUES
+
+
+class DataLoadingMixin:
+    """Mixin providing CSV data loading, grouping, and sign classification helpers."""
+
+    @staticmethod
+    def _point_from_csv_row(csv_row: dict) -> Point | None:
```
shundread, line 12: The thing that this method does is replicated in many parts of the commit. Maybe check for `Point(` calls and see if  all they're doing is `Point(float(thing[CSVHeadersV2.coord_x]), float(thing[CSVHeadersV2.coord_y]), 0.0)` and in those cases use this instead? I've marked three cases but I there could be more.

Maybe a couple of cases could be false flags, I don't know my brain is starting to turn to mush.

Some cases also care about the `coord_z` instead of assuming it's `0`.

traffic_control/analyze_utils/traffic_sign_data_v2_db_builders.py
```diff
@@ -0,0 +1,301 @@
...
+    def _build_additional_sign_reals_legacy_codes_by_source_id() -> dict[str, str | None]:
+        """Build mapping from source_id to device_type.legacy_code for AdditionalSignReal objects.
+
+        Returns:
+            dict[str, str | None]: Dictionary mapping source_id to device_type legacy_code.
+        """
+        return DbBuilderMixin._build_device_type_attr_map(AdditionalSignReal, "legacy_code")
+
+    @staticmethod
+    def _build_source_id_to_db_id(model_class: Type) -> dict[str, str]:
```
shundread, line 199: I'd rather see this used directly instead of several one-line specializations of it.

traffic_control/tests/management/commands/test_analyze_traffic_sign_data_v2.py
```diff
@@ -0,0 +1,1542 @@
...
+from traffic_control.analyze_utils.traffic_sign_data_v2_code_transform import CodeTransformMixin
+from traffic_control.models import MountType, TrafficControlDeviceType
+from traffic_control.tests.factories import SignpostRealFactory, TrafficSignRealFactory
+
+BASE_PATH = os.path.dirname(__file__)
+TEST_FILES_DIR = os.path.join(BASE_PATH, "../../test_datas/traffic_sign_import_v2")
+
+
+@pytest.fixture(autouse=True)
+def cleanup_filtered_csvs() -> None:
```
shundread: Use `tempfile.TemporaryDirectory` for a `test_files_dir` fixture instead of procedurally wiping the files?

traffic_control/tests/management/commands/test_analyze_traffic_sign_data_v2.py
```diff
@@ -0,0 +1,1542 @@
...
+    fields = (
+        ["1", obj_id]
+        + _COORDS.split(",")
+        + [mount_id, status]
+        + [code, "", "", "", "", numero, "", color, atsimuutti, parent_id, _TS, "3.0", ssurl]
+    )
+    return ",".join(fields) + "\n"
+
+
+def _write_csv(path: Path, header: str, rows: list[str]) -> str:
```
shundread: nitpick: use csv lib writer?

traffic_control/tests/management/commands/test_analyze_traffic_sign_data_v2.py
```diff
@@ -0,0 +1,1542 @@
...
+    """_point_from_csv_row returns a Point for valid coords or None for invalid.
+
+    Args:
+        x (str): X-coordinate string.
+        y (str): Y-coordinate string.
+        expected_none (bool): Whether None is expected.
+    """
+    row = {CSVHeadersV2.coord_x: x, CSVHeadersV2.coord_y: y}
+    result = TrafficSignAnalyzerV2._point_from_csv_row(row)
+    assert (result is None) is expected_none
```
shundread: Real weird way to write this. Why not

```python
if expected_none:
    assert result is None
else:
    assert isinstance(result, Point)
```

traffic_control/tests/management/commands/test_analyze_traffic_sign_data_v2.py
```diff
@@ -0,0 +1,1542 @@
...
+def test_group_by_mount_id() -> None:
+    """_group_by_mount_id groups sign dicts into lists keyed by mount_id."""
+    signs = {
+        "s1": {CSVHeadersV2.mount_id: "m1", CSVHeadersV2.id: "s1"},
+        "s2": {CSVHeadersV2.mount_id: "m1", CSVHeadersV2.id: "s2"},
+        "s3": {CSVHeadersV2.mount_id: "m2", CSVHeadersV2.id: "s3"},
+    }
+    result = TrafficSignAnalyzerV2._group_by_mount_id(signs)
+    assert {r[CSVHeadersV2.id] for r in result["m1"]} == {"s1", "s2"}
+    assert result["m2"][0][CSVHeadersV2.id] == "s3"
```
shundread, line 329: Also `assert length == 1` for `result["m2"]`?

traffic_control/tests/management/commands/test_analyze_traffic_sign_data_v2.py
```diff
@@ -0,0 +1,1542 @@
...
+
+    analyzer = _make_analyzer(tmp_path, [_mount_row("m1")], [_sign_row("obj1", "m1", "New", code)])
+
+    assert ("obj1" in analyzer.signposts_by_id) is in_signposts
+    assert ("obj1" in analyzer.signs_by_id) is in_signs
+    assert ("obj1" in analyzer.additional_signs_by_id) is in_additional
+
+
+@pytest.mark.django_db
+def test_skipable_exact_6_and_7_are_filtered(tmp_path: Path) -> None:
```
shundread: ~skipable~ -> skippable. Grep pr for other cases.

shundread (changes requested):
Still lots of fixes needed. Some clarifications wanted.

traffic_control/analyze_utils/traffic_sign_data_v2_code_transform.py
jforsman: Yeah.. there is no error case here, so no need to record any error. Different the _apply_number_code_replacement which actually check that the value is what it is supposed to be.

traffic_control/analyze_utils/traffic_sign_data_v2_code_transform.py
jforsman: It should record a failure.

traffic_control/analyze_utils/traffic_sign_data_v2_code_transform.py
jforsman: Valid point to be fixed, there was also hardcoded list of color values, that are now moved to contants file.

traffic_control/analyze_utils/traffic_sign_data_v2_db_builders.py
jforsman: it does? The line above obj.source_id: getattr(obj.device_type, attr) accesses the device_type attributes.

traffic_control/analyze_utils/traffic_sign_data_v2.py
jforsman: Added the missing wrapper method for signposts. I do not prefer your way, having the methods specially named makes it easier to read to me.

traffic_control/analyze_utils/traffic_sign_data_v2_db_builders.py
jforsman: I don't :D

traffic_control/analyze_utils/traffic_sign_data_v2_data_loading.py
jforsman: There was previously (in the first import run a year ago) some reason that we used 2d geometries for eg. distance calculations. In the reports and to db we will save the geometry as 3d. 
This still needs to be clarified with Esko.

traffic_control/tests/management/commands/test_analyze_traffic_sign_data_v2.py
jforsman: really should, does not even clean all currently

traffic_control/analyze_utils/traffic_sign_data_v2_db_builders.py
shundread: True. I was reading this line-by-line instead of reading it as a dict comprehension.

traffic_control/tests/management/commands/test_analyze_traffic_sign_data_v2.py
jforsman: Sure... can change

traffic_control/tests/management/commands/test_analyze_traffic_sign_data_v2.py
jforsman: Hmm... more maintainable incase the format changes so I will change it.

traffic_control/analyze_utils/traffic_sign_data_v2_reports.py
```diff
@@ -0,0 +1,735 @@
...
+            if not db_entry:
+                continue
+            db_id, db_location, *extra = db_entry
+            db_ssurl = extra[0] if extra else None
+            db_code = extra[1] if len(extra) > 1 else None
+            db_mount_type = extra[2] if len(extra) > 2 else None
+            csv_point = DataLoadingMixin._point_from_csv_row(csv_row)
+            if csv_point is None or db_location is None:
+                continue
+            db_point = Point(db_location.x, db_location.y, 0.0)
```
shundread, line 619: Missed one opportunity to use `_point_from_csv_row`

traffic_control/analyze_utils/traffic_sign_data_v2_constants.py
```diff
@@ -0,0 +1,209 @@
...
+}
+
+# Codes that should have location_specifier = 4
+LOCATION_SPECIFIER_4_CODES = ["4171", "4172", "418", "D3.1", "D3.1_2", "D3.2", "D3.2_2", "D3.3", "D3.3_2"]
+
+# Codes that should that are valid but should be skipped
+SKIPPABLE_CODES = ["6", "7"]
+
+# allowed color values
+ALLOWED_COLOR_VLUES = ["1", "2"]
```
shundread:
```diff
-ALLOWED_COLOR_VLUES
+ALLOWED_COLOR_VALUES
```

traffic_control/analyze_utils/traffic_sign_data_v2_data_loading.py
```diff
@@ -0,0 +1,262 @@
...
+        """Create a Point from CSV row coordinates.
+
+        Args:
+            csv_row (dict): CSV row dictionary containing coordinate fields.
+
+        Returns:
+            Point | None: A Point object, or None if coordinates are invalid.
+        """
+        try:
```
shundread, line 21: No try block necessary. Every usage of this in the code will eventually fail if `Point` resolves to `None` because they try to use `result.distance`. For example:

```python
            mount_point = self._point_from_csv_row(mount_data)
            data["distance_to_mount"] = mount_point.distance(self._point_from_csv_row(data))
```

And THOSE will crash anyway if `_point_from_csv_row` ever silences an exception.

traffic_control/analyze_utils/traffic_sign_data_v2.py
shundread: I don't care enough to argue this.

shundread (approved):
Tidy up a few things. Approving assuming no other big changes show up.

traffic_control/analyze_utils/traffic_sign_data_v2_reports.py
jforsman: Well.. not really, that is coming from db not from csv

traffic_control/analyze_utils/traffic_sign_data_v2_data_loading.py
jforsman: Yeah... this was a kinda of a known issue, probably aint gonna do nothing about it, if they fail then there really is something wrong with the data, badly and it should fail.

PR #844, opened 2026-05-04
feat: add signpost parent FKs to AdditionalSign and fix cascade-delet…

jforsman:
…ed signs

AdditionalSignPlan and AdditionalSignReal gain two new optional FK fields (signpost_plan / signpost_real) so that additional signs can reference either a TrafficSign or a Signpost as their parent. This resolves the issue where move_traffic_signs_to_signposts cascade-soft-deleted child additional signs.

Changes:
- Migration 0112: add signpost_plan/signpost_real FKs to additional sign models
- additional_sign.py: expose new FK fields; update clean() to accept signpost parent as valid alternative to traffic sign parent
- signpost.py / traffic_sign.py: minor related model adjustments
- move_traffic_signs_to_signposts: re-parent additional sign children before soft-deleting the original traffic sign
- fix_additional_signs_after_signpost_migration: new management command to restore additional signs that were cascade-soft-deleted by a prior migration run and re-parent them to the corresponding SignpostPlan/SignpostReal
- cleanup_migrated_traffic_signs: new management command to hard-delete TrafficSignPlan/TrafficSignReal objects soft-deleted by signpost or ticket-machine migration runs, with blocked-by-dependents protection

Refs: LIIK-952, LIIK-956


traffic_control/models/traffic_sign.py
```diff
@@ -256,6 +256,11 @@ class Meta:
             ),
         ]
 
+    @transaction.atomic
+    def soft_delete(self, user):
+        super(AbstractTrafficSign, self).soft_delete(user)
```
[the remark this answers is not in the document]
shundread: Github's suggestion is dumb.

Remove these `soft_delete` overrides entirely and just update the `soft_delete` method in `AbstractTrafficSign` since the implementation is exactly the same in both derived classes.

traffic_control/models/traffic_sign.py
```diff
@@ -379,6 +384,11 @@ class Meta:
...
```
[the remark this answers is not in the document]
shundread: Same comment about the suggested correction above

traffic_control/tests/management/commands/test_fix_additional_signs_after_signpost_migration.py
```diff
@@ -0,0 +1,470 @@
...
+User = get_user_model()
+
+CMD = "fix_additional_signs_after_signpost_migration"
+
+
+# ── Fixtures ───────────────────────────────────────────────────────────────────
+
+
+@pytest.fixture
+def system_user(db):
```
shundread, line 31: Implement with `get_system_user` from `users/utils.py` instead?

traffic_control/tests/management/commands/test_cleanup_migrated_traffic_signs.py
```diff
@@ -0,0 +1,654 @@
...
+)
+
+User = get_user_model()
+
+
+# ── Fixtures ──────────────────────────────────────────────────────────────────
+
+
+@pytest.fixture
+def system_user(db):
```
shundread, line 39: Implement with `get_system_user` from `users/utils.py` instead?

traffic_control/management/commands/fix_additional_signs_after_signpost_migration.py
```diff
@@ -0,0 +1,301 @@
...
+        Args:
+            victim: The soft-deleted AdditionalSign instance to restore.
+            new_signpost_obj: The SignpostPlan or SignpostReal to set as the new parent.
+            config: Type-specific configuration.
+            dry_run: If True, return True without saving.
+
+        Returns:
+            bool: True if restored (or would be in dry-run), False if already active.
+        """
+        if victim.is_active:
```
shundread: I'm not going to require a fix here since this doesn't break stuff, but I'm just curious about when is this condition true?

* _find_victims already deliberately filters by `"is_active": False`.
* _find_victims's query doesn't look to me like it would return duplicates, or if does we can just do `.unique()` there instead.
* If an object has been soft-deleted by a run, restored, and soft-deleted by another run, then by the time the fix for the second run occurs the object will no longer show up in the list of victims because `is_active=True`

Do I just misunderstand this?

traffic_control/management/commands/cleanup_migrated_traffic_signs.py
```diff
@@ -0,0 +1,288 @@
...
+        real_ids = set(reals_qs.values_list("id", flat=True))
+        plan_ids = set(plans_qs.values_list("id", flat=True))
+
+        deleted, _ = reals_qs.filter(id__in=real_ids - blocked_real_ids).delete()
+        logger.info("Hard-deleted %d TrafficSignReal objects", deleted)
+
+        deleted, _ = plans_qs.filter(id__in=plan_ids - blocked_plan_ids).delete()
+        logger.info("Hard-deleted %d TrafficSignPlan objects", deleted)
+
+    def _find_blocked_ids(self, ids: set, dependents: list[tuple[type, str]]) -> set:
```
shundread: Should this be in the collection helpers section?

traffic_control/tests/management/commands/test_cleanup_migrated_traffic_signs.py
```diff
@@ -0,0 +1,654 @@
...
+    Returns:
+        None
+    """
+    plan = TrafficSignPlanFactory(device_type=ticket_machine_device_type)
+    plan_id = plan.id
+    _soft_delete_plan(plan, system_user)
+
+    run = _make_ticket_machine_run(system_user)
+    new_id = uuid.uuid4()
+    TicketMachineMigrationPlanRecord.objects.create(
```
shundread, line 173: Would a pattern of calling the ticket machine migration command and getting the new ID from its records be better than replicating some steps of the migration? Genuine question, I don't actually know it myself but it's the feeling I got.

If so apply it to a bunch of tests below too.

traffic_control/tests/management/commands/test_cleanup_migrated_traffic_signs.py
```diff
@@ -0,0 +1,654 @@
...
+        migration_run=hard_delete_run,
+        original_id=plan_id,
+        new_id=uuid.uuid4(),
+        device_type_code=ticket_machine_device_type.code,
+    )
+
+    call_command("cleanup_migrated_traffic_signs", stdout=StringIO())
+
+    # Plan is still present (soft-deleted but not hard-deleted by cleanup)
```
shundread, line 338: How is this scenario possible? A failed hard-delete migration run + user-initiated soft-delete at a later point?

traffic_control/tests/management/commands/test_cleanup_migrated_traffic_signs.py
```diff
@@ -0,0 +1,654 @@
...
+        original_id=real_id,
+        new_id=uuid.uuid4(),
+        device_type_code=ticket_machine_device_type.code,
+    )
+
+    out = StringIO()
+    call_command("cleanup_migrated_traffic_signs", stdout=out)
+
+    assert TrafficSignReal.objects.filter(id=real_id).exists(), "Real should NOT be deleted — it has a child"
+    assert additional.__class__.objects.filter(id=additional.id).exists(), "AdditionalSignReal must be untouched"
```
shundread: wtf

`AdditionalSignReal.objects.filter`

traffic_control/tests/management/commands/test_cleanup_migrated_traffic_signs.py
```diff
@@ -0,0 +1,654 @@
...
+        original_id=plan_id,
+        new_id=uuid.uuid4(),
+        device_type_code=ticket_machine_device_type.code,
+    )
+
+    out = StringIO()
+    call_command("cleanup_migrated_traffic_signs", stdout=out)
+
+    assert TrafficSignPlan.objects.filter(id=plan_id).exists(), "Plan should NOT be deleted — it has a child"
+    assert additional.__class__.objects.filter(id=additional.id).exists(), "AdditionalSignPlan must be untouched"
```
shundread: `AdditionalSignPlan.objects.filter`

traffic_control/tests/management/commands/test_fix_additional_signs_after_signpost_migration.py
```diff
@@ -0,0 +1,470 @@
...
+    Returns:
+        None
+    """
+    traffic_sign_plan = TrafficSignPlanFactory()
+    signpost_plan = SignpostPlanFactory()
+    additional = AdditionalSignPlanFactory(parent=traffic_sign_plan)
+    _soft_delete_sign(additional, system_user, completed_at)
+
+    run = _make_signpost_run(system_user, completed_at)
+    SignpostMigrationPlanRecord.objects.create(
```
shundread, line 114: Same question about running the command vs recreating some of its steps.

traffic_control/tests/management/commands/test_fix_additional_signs_after_signpost_migration.py
```diff
@@ -0,0 +1,470 @@
...
+    AdditionalSignRealFactory,
+    SignpostPlanFactory,
+    SignpostRealFactory,
+    TrafficSignPlanFactory,
+    TrafficSignRealFactory,
+)
+
+User = get_user_model()
+
+CMD = "fix_additional_signs_after_signpost_migration"
```
shundread, line 24: Not an error, but inconsistent style. Why the distinct pattern of using a CMD constant here vs repeating the name of the command in each test for the other command?

traffic_control/models/signpost.py
```diff
@@ -34,6 +35,39 @@
 from traffic_control.models.mount import MountPlan, MountReal
 from traffic_control.models.plan import Plan
 from traffic_control.models.traffic_sign import LocationSpecifier
+from traffic_control.models.utils import SoftDeleteQuerySet
+
+
+class SignpostPlanQuerySet(SoftDeleteQuerySet):
+    """QuerySet for SignpostPlan that cascades soft-delete to additional signs."""
+
+    def soft_delete(self, user: object) -> None:
```
shundread, line 44: Add test that this soft_delete behaves as expected, also should it be atomic? Review these needs for the other newly defined soft_delete in the pull request.

shundread (changes requested):
mostly good

traffic_control/management/commands/fix_additional_signs_after_signpost_migration.py
jforsman: it is never True, yeah.. too defensive, I will probably remove it

traffic_control/management/commands/cleanup_migrated_traffic_signs.py
jforsman: yes

traffic_control/tests/management/commands/test_cleanup_migrated_traffic_signs.py
jforsman: I will rather keep it like this, would just make the test more complicated, would still need to setup db entries to run the migration.

traffic_control/tests/management/commands/test_cleanup_migrated_traffic_signs.py
jforsman: wtf to your "wtf"? :D

traffic_control/tests/management/commands/test_fix_additional_signs_after_signpost_migration.py
jforsman: Same comment as on the previous similar comment, keeping it like this.

traffic_control/tests/management/commands/test_cleanup_migrated_traffic_signs.py
jforsman: Well, user initiated soft-delete cannot happen after hard-delete. I get your point that this is almost useless test, but it just tests that hard-delete runs are ignored. I can remove the whole test if you insist on it :D

traffic_control/tests/management/commands/test_cleanup_migrated_traffic_signs.py
jforsman: Ah.. the __class__ part

traffic_control/tests/management/commands/test_fix_additional_signs_after_signpost_migration.py
jforsman: not inconsistent within this file, this is better ofc than repeating the command name, right? Change the cleanup tests to use the same style,

traffic_control/management/commands/fix_additional_signs_after_signpost_migration.py
```diff
@@ -0,0 +1,298 @@
...
+                config.parent_fk_field: original_id,
+                "is_active": False,
+                f"{config.signpost_fk_field}__isnull": True,
+                "deleted_at__gte": window_start,
+                "deleted_at__lte": window_end,
+            }
+        )
+
+    @transaction.atomic
+    def _restore_victim(
```
shundread, line 255: I think this SonarQube complaint is worth taking a look at. `_restore_victim` now returns always True, so that means it doesn't need to return anything at all. It also means that the script no longer needs a concept of "skipped" elements.

* `_restore_victim`
  * doesn't need return value, only early return for `dry_run` case.
* `_fix_for_type`
  * doesn't need a "skipped" branch
  * doesn't need to count or return amount of skipped elements, it's always 0
* `handle`
  *  doesn't need to track skipped values
  * doesn't need to write skipped counts to stdout

It's a bit of work tidying this up, since it cuts down on a meaningful amount of code, but less code to maintain is better.

traffic_control/tests/test_signpost_soft_delete_cascade.py
```diff
@@ -0,0 +1,256 @@
+"""Tests for SignpostPlan and SignpostReal soft-delete cascade to AdditionalSign objects."""
```
shundread, line 1: In the tests where you assert that a `deleted_at` hasn't been affected by a cascade delete operation, can you also check that `deleted_by` hasn't been affected either?

Something like:

```diff
diff --git a/traffic_control/tests/test_signpost_soft_delete_cascade.py b/traffic_control/tests/test_signpost_soft_delete_cascade.py
index 2e6bf20..9339971 100644
--- a/traffic_control/tests/test_signpost_soft_delete_cascade.py
+++ b/traffic_control/tests/test_signpost_soft_delete_cascade.py
@@ -7,6 +7,7 @@ from traffic_control.tests.factories import (
     AdditionalSignRealFactory,
     SignpostPlanFactory,
     SignpostRealFactory,
+    UserFactory,
 )
 from users.utils import get_system_user
 
@@ -26,6 +27,11 @@ def system_user(db):
     return get_system_user()
 
 
+@pytest.fixture
+def other_user(db):
+    return UserFactory(username="other", first_name="Other", last_name="User", email="other@example.com")
+
+
 # ── SignpostPlan instance soft_delete ──────────────────────────────────────────
 
 
@@ -52,7 +58,7 @@ def test_signpost_plan_soft_delete_cascades_to_additional_sign_plan(system_user)
 
 
 @pytest.mark.django_db
-def test_signpost_plan_soft_delete_does_not_overwrite_already_inactive_additional_sign_plans(system_user):
+def test_signpost_plan_soft_delete_does_not_overwrite_already_inactive_additional_sign_plans(system_user, other_user):
     """Already-inactive AdditionalSignPlan children are not re-deleted and their deleted_at is preserved.
 
     Args:
@@ -64,8 +70,9 @@ def test_signpost_plan_soft_delete_does_not_overwrite_already_inactive_additiona
     signpost_plan = SignpostPlanFactory()
     active_additional = AdditionalSignPlanFactory(signpost_plan=signpost_plan, parent=None)
     inactive_additional = AdditionalSignPlanFactory(signpost_plan=signpost_plan, parent=None)
-    inactive_additional.soft_delete(system_user)
+    inactive_additional.soft_delete(other_user)
     original_deleted_at = inactive_additional.deleted_at
+    original_deleted_by = inactive_additional.deleted_by
 
     signpost_plan.soft_delete(system_user)
 
@@ -73,6 +80,7 @@ def test_signpost_plan_soft_delete_does_not_overwrite_already_inactive_additiona
     inactive_additional.refresh_from_db()
     assert active_additional.is_active is False
     assert inactive_additional.deleted_at == original_deleted_at
+    assert inactive_additional.deleted_by == original_deleted_by
```

shundread (changes requested):
Two last fixes, otherwise looks good.

traffic_control/management/commands/fix_additional_signs_after_signpost_migration.py
shundread: "this SonarQube complaint" -> https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=844&id=City-of-Helsinki_city-infrastructure-platform&open=AZ34N9RN82LJpwge5JhG

traffic_control/tests/test_signpost_soft_delete_cascade.py
jforsman: sure... can add this

traffic_control/serializers/common.py
```diff
@@ -105,41 +105,95 @@ class AdditionalSignParentValidationMixin:
...
+        Returns:
+            The device type to use for validation.
+        """
+        device_type = data.get("device_type")
+        if is_update and not device_type:
+            return self.instance.device_type
+        return device_type
+
+    @staticmethod
+    def _raise_parent_required_error() -> None:
```
shundread, line 157: This doesn't look like a simplification at all, I now have to skip back and forth to see what one single statement does, and this is used in total once in the whole codebase, so this isn't a deduplication either.

Are you genAI settings requesting minimal edits? LLMs tend to refactor a bunch of unrelated stuff every time you're trying to fix a small thing with them. I've gotten into the habit of adding `don't change unrelated stuff` to my prompts and manually tweaking / cutting away part of what's in the suggested diff.

The localized text got altered, so so you should either re-run `makemessages.sh` or restore the old

```python
            if device_type.code not in TICKET_MACHINE_CODES:
                raise serializers.ValidationError(
                    {"parent": _("Parent is required for additional signs that are not ticket machines.")}
                )
```

shundread (approved):
Requesting one change but I trust that you'll fix this cleanly. Feel free to merge after the fix.

traffic_control/serializers/common.py
jforsman: Welll... if want to start making exceptions to sonarqube analysis criterieas go ahead or make a real suggestion what to change and how and why. 
Will do the translations changes. Probably some more coming than just this....

jforsman: It was no unrelated, it was done because sonarqube cognitive complexity was over the set limit.

PR #845, opened 2026-05-05
feat: Add is_active to metadata section for admin models where applic…

jforsman:
…able

Refs: LIIK-955


shundread (approved):
Good.

The AI-generated complaint about code duplication is garbage.

PR #856, opened 2026-05-20
chore: Update outated python, node, uv, github actions

shundread:
Github actions pinned to release SHAs

Refs: LIIK-921

### DIFF
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -8,30 +8,36 @@ on:
 
 jobs:
   build:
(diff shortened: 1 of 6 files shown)


jforsman:
Looks good, better to have those immutable commit sha1 checksums instead of the version in the action.
But what about the python version... does it need some update to the devops repo also? And update when running things locally? If not then it is ok.

jforsman (approved):
devops python version and the actual version used when running pods is the only question.

PR #857, opened 2026-05-20
feat: Filter out rows from duplicate sings on same mount report

jforsman:
* no legacy code matching for the no-exact code report

Refs: LIIK-961

### DIFF
--- a/traffic_control/analyze_utils/traffic_sign_data_v2_constants.py
+++ b/traffic_control/analyze_utils/traffic_sign_data_v2_constants.py
@@ -127,6 +127,10 @@ CODE_AND_COLOR_DEPENDENT_CODES = {
     "833_2": {"new_code": "833_2", "color_1_suffix": "S", "color_2_suffix": None},
 }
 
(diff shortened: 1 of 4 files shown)


traffic_control/analyze_utils/traffic_sign_data_v2_constants.py
```diff
@@ -127,6 +127,10 @@
     "833_2": {"new_code": "833_2", "color_1_suffix": "S", "color_2_suffix": None},
 }
 
+# Codes excluded from duplicate-on-same-mount reports when the duplicate count is exactly 2
+# (e.g. double-sided signs that legitimately appear in pairs)
+DUPLICATE_PAIR_EXCLUDED_CODES: frozenset[str] = frozenset(["5111", "5112", "E1", "E1_2", "5311", "5331", "E6", "E7"])
```
shundread, line 132: This is the only usage of `frozenset` in the entire codebase. Nothing against using a read-only object for const objects. Actually a good practice I suppose, but why only in this place?

shundread (approved):
One small comment

PR #864, opened 2026-05-27
feat: Streetscan2025 data import, skeleton and mounts

jforsman:
* Management command parameter handling
* Import functionality for MountReal create and update
* Skeleton implementation for the rest of the coming features

LIIK-964


traffic_control/models/streetscan_import.py
```diff
@@ -0,0 +1,182 @@
...
+class StreetScanImportRun(models.Model):
+    """Tracks a single execution of the import_streetscan_signs_v2 management command.
+
+    Each live or dry-run invocation writes one row here, recording counts,
+    skip/warning/error details, and the revert data file attachment.
+    The revert file (if any) is stored as a related StreetScanImportRevertFile record.
+    """
+
+    id = models.BigAutoField(primary_key=True)
+    ran_at = models.DateTimeField(
```
shundread: Replace with `started_at` and `completed_at` so it's similar to `SignpostMigrationRun`

traffic_control/models/streetscan_import.py
```diff
@@ -0,0 +1,182 @@
...
+        _("Ran at"),
+        auto_now_add=True,
+        help_text=_("Timestamp when the import run was started."),
+    )
+    is_dry_run = models.BooleanField(
```
shundread: Every other command that takes `--dry-run` option and logs it uses `dry_run` instead of `is_dry_run`

traffic_control/models/streetscan_import.py
```diff
@@ -0,0 +1,182 @@
...
+    )
+    error_count = models.IntegerField(
+        _("Error count"),
+        default=0,
+        help_text=_("Total error-level events (unexpected exceptions during the run)."),
+    )
+
+    # --- Timing ---
+
+    preprocessing_duration_s = models.FloatField(
```
shundread: Is there a reason why this isn't simply added to the set of `phase_durations` below?

traffic_control/admin/streetscan_import.py
```diff
@@ -0,0 +1,458 @@
...
+            _("Phase Durations"),
+            {
+                "fields": ("phase_durations_display",),
+                "classes": ("collapse",),
+            },
+        ),
+    )
+    inlines = [StreetScanImportRevertFileInline]
+
+    def has_add_permission(self, request) -> bool:
```
shundread, line 214: For stuff that's simply an override of methods that through internal django processes produce magic results, I'd simply link to the reference of the base class' method and clarify in a brief block instead of a the big block of documentation for a method we don't ever directly call:

```diff
+        # Disable creating run logs through admin
+        # https://docs.djangoproject.com/en/5.2/ref/contrib/admin/#django.contrib.admin.ModelAdmin.has_add_permission
-        """Disable creating run logs through admin.
-        (...)
-        """
```

Same advice for the other similar override in this class.

traffic_control/admin/streetscan_import.py
```diff
@@ -0,0 +1,458 @@
...
+            obj (StreetScanImportRun): The run log instance.
+
+        Returns:
+            str: Safe HTML table of object-type × phase durations.
+        """
+        durations: dict = obj.phase_durations or {}
+        if not durations:
+            return mark_safe('<span style="color:#6c757d;">No timing data recorded.</span>')
+
+        rows = "".join(
```
shundread: By the time you're doing multi-step concatenation of stuff with a mix of `join`, `format_html` and `mark_safe` it's better to use a template + `render_to_string` instead.

See `PlanGeometryImportLogAdmin._format_results`

Same advice for `details_display` below.

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
```diff
@@ -0,0 +1,1528 @@
+"""V2 traffic sign importer.
+
+Handles create, update and deactivate operations for MountReal, TrafficSignReal,
+SignpostReal and AdditionalSignReal based on enriched V2 CSV data.
+
+This module is intentionally kept separate from the analysis pipeline so that
+the management command can run an import without triggering the full analysis.
+"""
+from __future__ import annotations
```
shundread: ???

https://docs.python.org/3/library/__future__.html#id2

> from `__future__ import annotations` was previously scheduled to become mandatory in Python 3.10, but the change was delayed and ultimately canceled. This feature will eventually be deprecated and removed. See [PEP 649](https://peps.python.org/pep-0649/) and [PEP 749](https://peps.python.org/pep-0749/).

remove this and the other cases of this in the PR

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
```diff
@@ -0,0 +1,1528 @@
...
+        # Normalise and sort by dependency order.
+        self.object_types: list[str] = [ot for ot in OBJECT_TYPE_ORDER if ot in object_types]
+        self.phases: list[str] = list(phases)
+
+        # --- CSV loading ---
+        _t_preprocess_start = datetime.datetime.now()
+        self.mount_rows: list[dict] = self._read_csv_file(mount_file, delimiter)
+        self.sign_rows: list[dict] = self._read_csv_file(sign_file, delimiter)
+
+        # --- Sign row enrichment (code transforms + internal status) ---
```
shundread, line 137: I'm thinking some or these phases could

1. be broken down into methods like `def group_rows_by_csv_id`, inside `TrafficSignAnalyzerV2` or some other utility Mixin
2. be called in its `analyze` method inside of running during `__init__`
3. as a result, imported into this class through inheritance
4. called here

We had a chat about it and you disagreed but I thought I'd leave the comment anyway. Feel free to say no and resolve.

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
```diff
@@ -0,0 +1,1528 @@
...
+
+        # --- Row grouping by CSV id ---
+        self.mounts_by_id: dict = self._get_objects_by_id(self.mount_rows)
+        self.all_signs_by_id: dict = self._get_objects_by_id(self.sign_rows)
+        self.signs_by_id: dict = self._get_signs_by_id(self.sign_rows)
+        self.additional_signs_by_id: dict = self._get_additional_signs_by_id(self.sign_rows)
+        self.signposts_by_id: dict = self._get_signposts_by_id(self.sign_rows)
+
+        # --- Device type FK lookup ---
+        self.code_to_device_type_id: dict = self._build_code_to_device_type_mapping()
```
shundread, line 163: Some of the steps leading up to this method call in this class happen later than the call to this method in `TrafficSignAnalyzerV2`. Does it matter?

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
```diff
@@ -0,0 +1,1528 @@
...
+        self._apply_summary_to_run_log(self.run_log, summary)
+        self.run_log.save()
+
+    def _create_run_log(self) -> StreetScanImportRun:
+        """Create and persist the initial StreetScanImportRun row for this run.
+
+        Returns:
+            StreetScanImportRun: The newly created (saved) run log instance.
+        """
+        run_log = StreetScanImportRun(
```
shundread:
```diff
- run_log = StreetScanImportRun(...)
- run_log.save()
+ run_log = StreetScanImportRun.objects.create(...)
```

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
```diff
@@ -0,0 +1,1528 @@
...
+        Args:
+            skip_source_ids (set[str]): Source IDs to exclude (already exist in DB,
+                i.e. keys of mount_source_id_to_db_id).
+            summary (dict[str, Any]): Mutable summary dict for skip/warning entries.
+            phase_started_at (datetime.datetime): Timestamp of phase start, used as created_at.
+
+        Yields:
+            MountReal: Unsaved MountReal instance ready for bulk_create.
+        """
+        mount_types_by_name: dict[str, MountType] = {
```
shundread: We're hitting the database twice here for the exact same query. A few similar cases exist. Grep for `mount_types_by_name: dict[str, MountType] = {`

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
```diff
@@ -0,0 +1,1528 @@
...
+            date_str (str | None): Raw timestamp string from CSV, or None.
+
+        Returns:
+            datetime.datetime | None: Parsed UTC-aware datetime, or None if unparseable.
+        """
+        if not date_str:
+            return None
+        try:
+            return datetime.datetime.strptime(date_str.strip() + "00", "%Y/%m/%d %H:%M:%S%z")
+        except ValueError:
```
shundread, line 529: Report error?

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
```diff
@@ -0,0 +1,1528 @@
...
+from traffic_control.models import MountReal, MountType, Owner, TrafficSignReal
+from traffic_control.models.mount import LocationSpecifier as MountLocationSpecifier
+from traffic_control.models.streetscan_import import StreetScanImportRevertFile, StreetScanImportRun
+from traffic_control.models.traffic_sign import LocationSpecifier as SignLocationSpecifier
+from users.models import User
+
+logger = logging.getLogger(__name__)
+
+# Valid values for the --object-type and --phase CLI arguments.
+VALID_OBJECT_TYPES: tuple[str, ...] = ("mounts", "signs", "signposts", "additional-signs")
```
shundread, line 42: Is it safe to replace all cases of `"additional-signs"` with `"additional_signs"` or are we getting the `"additional-signs"` string somewhere in the data?

Nowhere else in the codebase we write it like this.

```
$ git grep -c additional-signs
cityinfra/urls.py:1
traffic_control/analyze_utils/traffic_sign_data_v2_import.py:6
traffic_control/tests/management/commands/test_import_streetscan_signs_v2.py:2
```

Even in this same file we later use `"additional_signs_created"`

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
shundread, line 1: Reviewing this file I got the very strong impression that you should be using `defafaultdict` objects in many places instead of `dict.get` with fallbacks.

```
$ git grep "summary.get("
traffic_control/analyze_utils/traffic_sign_data_v2_import.py:        phase_result = summary.get("phase_results", {}).get(object_type, {}).get(phase)
traffic_control/analyze_utils/traffic_sign_data_v2_import.py:        details_before = len(summary.get("details", []))
traffic_control/analyze_utils/traffic_sign_data_v2_import.py:        new_details = summary.get("details", [])[details_before:]
traffic_control/analyze_utils/traffic_sign_data_v2_import.py:        details: list[dict] = summary.get("details", [])
traffic_control/analyze_utils/traffic_sign_data_v2_import.py:        for obj_type, phases in summary.get("phase_results", {}).items():
traffic_control/analyze_utils/traffic_sign_data_v2_import.py:        details_before = len(summary.get("details", []))
traffic_control/analyze_utils/traffic_sign_data_v2_import.py:        new_details = summary.get("details", [])[details_before:]
traffic_control/analyze_utils/traffic_sign_data_v2_import.py:        details_before = len(summary.get("details", []))
traffic_control/analyze_utils/traffic_sign_data_v2_import.py:        new_details = summary.get("details", [])[details_before:]
traffic_control/management/commands/import_streetscan_signs_v2.py:        self.stdout.write(f"  object_types : {summary.get('object_types')}")
traffic_control/management/commands/import_streetscan_signs_v2.py:        self.stdout.write(f"  phases       : {summary.get('phases')}")
traffic_control/management/commands/import_streetscan_signs_v2.py:        self.stdout.write(f"  dry_run      : {summary.get('dry_run')}")
traffic_control/management/commands/import_streetscan_signs_v2.py:        phase_results: dict = summary.get("phase_results", {})
traffic_control/management/commands/import_streetscan_signs_v2.py:        details: list = summary.get("details", [])
traffic_control/tests/management/commands/test_traffic_sign_importer_v2_create_mounts.py:    result = summary.get("phase_results", {}).get("mounts", {}).get("create")
traffic_control/tests/management/commands/test_traffic_sign_importer_v2_create_mounts.py:    assert "PROC1" in summary.get("processed_mount_source_ids", [])
traffic_control/tests/management/commands/test_traffic_sign_importer_v2_create_mounts.py:    assert "PROC2" in summary.get("processed_mount_source_ids", [])
traffic_control/tests/management/commands/test_traffic_sign_importer_v2_create_mounts.py:    assert "SKIP_PROC" not in summary.get("processed_mount_source_ids", [])
traffic_control/tests/management/commands/test_traffic_sign_importer_v2_signs.py:    result = summary.get("phase_results", {}).get("signs", {}).get("create")
traffic_control/tests/management/commands/test_traffic_sign_importer_v2_signs.py:    result = summary.get("phase_results", {}).get("signs", {}).get("update")
traffic_control/tests/management/commands/test_traffic_sign_importer_v2_signs.py:    result = summary.get("phase_results", {}).get("signs", {}).get("deactivate")
traffic_control/tests/management/commands/test_traffic_sign_importer_v2_update_mounts.py:    result = summary.get("phase_results", {}).get("mounts", {}).get("update")
```

Particularly lines like these suck:

```
result = summary.get("phase_results", {}).get("mounts", {}).get("update")
```

It feels a bit like `row[col].strip()` showing up everywhere when it really only needed to be done in a single place, but for the `summary` object is a bit trickier since the default value we're expecting can be `0`, `{}` or `[]`, but if `summary` is split into three (?) objects it would tidy up this stuff a lot.

This sort of change might demand wider fixes on snippets like this:

```python
        for field in (
            "mounts_created",
            "mounts_updated",
            "signs_created",
            "signs_updated",
            "signs_deactivated",
            "signposts_created",
            "signposts_updated",
            "signposts_deactivated",
            "additional_signs_created",
            "additional_signs_updated",
            "additional_signs_deactivated",
            "processed_mount_source_ids",
            "processed_sign_source_ids",
            "processed_signpost_source_ids",
            "processed_additional_sign_source_ids",
        ):
            if field in summary:
                setattr(run_log, field, summary[field])
```

But could still dbe done with a pattern like `for field in FIELDSET_1: ...` + `for field in FIELDSET_2: ...`

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
```diff
@@ -0,0 +1,1528 @@
...
+                logger.debug("_attach_revert_file: opened tmp for reading, calling file.save()")
+                revert_record.file.save(filename, File(fh), save=True)
+            logger.debug(
+                "_attach_revert_file: saved — storage name=%s, pk=%s",
+                revert_record.file.name,
+                revert_record.pk,
+            )
+        finally:
+            try:
+                os.unlink(tmp_path)
```
shundread, line 666: Something about having a temporary file that needs to be explicitly deleted smells. How big does the file end up looking like? Is it possible to simply use a big string instead and `json.dumps`?

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
```diff
@@ -0,0 +1,1528 @@
...
+            mount_types_by_name (dict[str, MountType]): MountType lookup by name.
+            summary (dict[str, Any]): Mutable summary dict; skips are appended to
+                summary["details"] and summary["skipped_mount_update_count"] is
+                incremented.
+            phase_started_at (datetime.datetime): Timestamp of phase start, used as updated_at.
+
+        Yields:
+            MountReal: Mutated (unsaved) MountReal instance ready for bulk_update.
+        """
+        skipped: list[int] = summary.setdefault("_skipped_mount_update_count", [0])
```
shundread: Why is this a list with a single number instead of just being the number?

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
```diff
@@ -0,0 +1,1528 @@
...
+            height_str (str | None): Raw height string from CSV, or None.
+
+        Returns:
+            int | None: Height in centimetres, or None if unparseable.
+        """
+        if not height_str:
+            return None
+        try:
+            return int(float(height_str) * 100)
+        except (ValueError, TypeError):
```
shundread, line 1002: report error?

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
```diff
@@ -0,0 +1,1528 @@
...
+            direction_str (str | None): Raw direction string from CSV, or None.
+
+        Returns:
+            int | None: Direction in degrees, or None if unparseable.
+        """
+        if not direction_str:
+            return None
+        try:
+            return int(float(direction_str))
+        except (ValueError, TypeError):
```
shundread, line 1024: report error?

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
```diff
@@ -0,0 +1,1528 @@
...
+            condition_str (str | None): Raw condition string from CSV, or None.
+
+        Returns:
+            Condition | None: Parsed Condition enum value, or None if unparseable.
+        """
+        if not condition_str:
+            return None
+        try:
+            return Condition(int(condition_str))
+        except (ValueError, TypeError):
```
shundread, line 1046: report error?

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
```diff
@@ -0,0 +1,1528 @@
...
+            Decimal | None: Extracted value, or None if no numeric prefix found.
+        """
+        if not number_code_str:
+            return None
+        match = NUMBER_CODE_PATTERN.match(number_code_str.strip())
+        if not match:
+            return None
+        try:
+            return Decimal(match.group(1))
+        except Exception:
```
shundread, line 1085: report error?

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
```diff
@@ -0,0 +1,1528 @@
...
+            phase_started_at (datetime.datetime): Timestamp of phase start, used as created_at.
+
+        Yields:
+            TrafficSignReal: Unsaved TrafficSignReal instance ready for bulk_create.
+        """
+        default_owner = self._get_default_owner()
+        try:
+            private_owner = Owner.objects.get(name_fi="Yksityinen")
+        except Owner.DoesNotExist:
+            private_owner = default_owner
```
shundread: Is this an okay outcome to case where the owner does not exist? Does it need some error reporting?

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
```diff
@@ -0,0 +1,1528 @@
...
+                appended to summary["details"].
+            phase_started_at (datetime.datetime): Timestamp of phase start, used as updated_at.
+
+        Yields:
+            TrafficSignReal: Mutated (unsaved) instance ready for bulk_update.
+        """
+        skipped: list[int] = summary.setdefault("_skipped_sign_update_count", [0])
+        details: list[dict] = summary.setdefault("details", [])
+
+        for source_id in update_source_ids:
```
shundread: A lot of the logic here is duplicated from `_get_signs_to_update`, probably also a big intersection with `_get_mounts_to_update`. It's just not triggering code duplication complaints because of minor string differences.

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
```diff
@@ -0,0 +1,1528 @@
...
+        }
+        mount_types_by_name: dict[str, MountType] = {
+            **{mt.description_fi: mt for mt in MountType.objects.all()},
+            **{mt.description: mt for mt in MountType.objects.all()},
+        }
+        default_owner = self._get_default_owner()
+        try:
+            private_owner = Owner.objects.get(name_fi="Yksityinen")
+        except Owner.DoesNotExist:
+            private_owner = default_owner
```
shundread: Again the question if this exception is ok to silence

traffic_control/tests/management/commands/test_import_streetscan_signs_v2.py
```diff
@@ -0,0 +1,527 @@
...
+    VALID_OBJECT_TYPES,
+    VALID_PHASES,
+)
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+_MOUNT_CSV_HEADER = ["OBJECTID", "id", "x", "y", "z", "stdx", "stdy", "stdz", "status", "tallennusajankohta", "ssurl"]
+_SIGN_CSV_HEADER = [
```
shundread, line 16: These two constants look like they belong in application code

traffic_control/models/streetscan_import.py
jforsman: True, will change it. phase_durations is not even working currently, not shown in admin page.

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
jforsman: This is on my "ask from Esko" list

traffic_control/tests/management/commands/test_import_streetscan_signs_v2.py
```diff
@@ -0,0 +1,527 @@
...
+    assert "Mount file not found" in stderr
+
+
+# ===========================================================================
+# Defaults: omitting --object-type and --phase
+# ===========================================================================
+
+
+@pytest.mark.django_db
+def test_default_object_types_and_phases(tmp_path: Path) -> None:
```
shundread: Why do we have a bunch of tests that seem to be just testing argparse? There's ~250 lines of test of a library very well known to behave well. This is not ok.

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
jforsman: But it really should not happen at all, maybe we need to add some pre-preprocessing phase to check that the db has all expected entries, also the helsinki owner

traffic_control/tests/management/commands/test_import_streetscan_signs_v2.py
```diff
@@ -0,0 +1,527 @@
...
+    Args:
+        tmp_path (Path): Pytest tmp_path fixture.
+    """
+    mount_file, _ = _make_csv_files(tmp_path)
+    stdout, stderr = _call(mount_file, "/nonexistent/signs.csv")
+    assert "Sign file not found" in stderr
+    assert stdout == ""
+
+
+def test_both_files_missing_reports_mount_error_first(tmp_path: Path) -> None:
```
shundread, line 133: What is the value of this test? If either one is missing we get an error and that's sufficient, and that's covered by the two tests above.

traffic_control/tests/management/commands/test_import_streetscan_signs_v2.py
```diff
@@ -0,0 +1,527 @@
...
+        mock_cls.return_value = mock_instance
+
+        stdout, _ = _call(mount_file, sign_file)
+
+        assert "bad_id_1" in stdout
+        assert "device type not found" in stdout
+
+
+@pytest.mark.django_db
+def test_skip_and_warning_counts_in_summary(tmp_path: Path) -> None:
```
shundread: The test is named "test skip and warning **counts** in summary" but it never looks at the counts.

traffic_control/tests/management/commands/test_import_streetscan_signs_v2.py
```diff
@@ -0,0 +1,527 @@
...
+        assert "warnings" in stdout
+
+
+# ===========================================================================
+# Importer is called exactly once
+# ===========================================================================
+
+
+@pytest.mark.django_db
+def test_importer_run_called_once(tmp_path: Path) -> None:
```
shundread: Useless test.

traffic_control/tests/management/commands/test_import_streetscan_signs_v2.py
shundread, line 1: There isn't much value in this test file.

shundread (changes requested):

traffic_control/tests/management/commands/test_traffic_sign_importer_v2_additional_signs.py
```diff
@@ -0,0 +1,1095 @@
...
+
+@pytest.mark.django_db
+def test_create_additional_signs_inserts_new_records(tmp_path: Path, default_owner, device_type, mount_type) -> None:
+    """New additional sign rows are inserted into the database.
+
+    Args:
+        tmp_path (Path): Pytest tmp_path fixture.
+        default_owner: Owner fixture.
+        device_type: DeviceType fixture.
+        mount_type: MountType fixture.
```
shundread, line 239: What's the value of these args docs? It's a pytest test, all the args are fixtures.

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
jforsman: We can discuss this more later, probably the last change that would come :D

traffic_control/tests/management/commands/test_traffic_sign_importer_v2_additional_signs.py
```diff
@@ -0,0 +1,1095 @@
...
+        location=Point(25497100.0, 6673400.0, 5.0, srid=3879),
+        parent=None,
+    )
+
+    importer = _make_importer(tmp_path, [_sign_row("ASU1")], phases=["update"])
+    summary: dict = {"details": []}
+    importer._update_additional_signs(summary)
+
+    obj = AdditionalSignReal.objects.get(source_id="ASU1")
+    assert obj.location.x == pytest.approx(25497188.0, abs=1)
```
shundread, line 573: Better to compare it with `_COORDS[0]`

traffic_control/tests/management/commands/test_traffic_sign_importer_v2_additional_signs.py
```diff
@@ -0,0 +1,1095 @@
...
+        default_owner: Owner fixture.
+        device_type: DeviceType fixture.
+        mount_type: MountType fixture.
+    """
+    AdditionalSignRealFactory(
+        source_id="ASU1",
+        source_name="StreetScan2025",
+        owner=default_owner,
+        device_type=device_type,
+        location=Point(25497100.0, 6673400.0, 5.0, srid=3879),
```
shundread, line 564: Better to define the coordinates in terms of `_COORDS[n] + delta_n`, may require switching `_COORDS` to `list[float]` type.

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
jforsman: Does not matter, it is only used to in the import phase to make the FK references in the objects that need device type. Resolving this.

traffic_control/tests/management/commands/test_traffic_sign_importer_v2_additional_signs.py
```diff
@@ -0,0 +1,1095 @@
...
+    summary: dict = {"details": []}
+    importer._update_additional_signs(summary)
+
+    obj = AdditionalSignReal.objects.get(source_id="ASU1")
+    assert obj.location.x == pytest.approx(25497188.0, abs=1)
+    assert summary["additional_signs_updated"] == 1
+
+
+@pytest.mark.django_db
+def test_update_additional_signs_sets_updated_at(tmp_path: Path, default_owner, device_type, mount_type) -> None:
```
shundread, line 578: This test does not correctly test that `updated_at` is affected by the importer.

```python
    additional = AdditionalSignRealFactory(
        source_id="ASUA",
        source_name="StreetScan2025",
        owner=default_owner,
        device_type=device_type,
        location=Point(25497100.0, 6673400.0, 5.0, srid=3879),
        parent=None,
    )
    assert additional.updated_at is None # <- It's going to fail here.
```

Same problem for the other similar test in this file.

shundread (changes requested):

traffic_control/tests/management/commands/test_import_streetscan_signs_v2.py
shundread: This pattern happens in other test files too.

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
jforsman: need to check all these _get_xxxx methods

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
jforsman: Yes it is safe, it is just the object type given from the management command basically.

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
jforsman: missing those owners from db is now an error that prevents running totally

traffic_control/tests/management/commands/test_import_streetscan_signs_v2.py
jforsman: Can remove them totally if you insist

traffic_control/tests/management/commands/test_import_streetscan_signs_v2.py
jforsman: Well... kinda yeah... will remove it.

traffic_control/tests/management/commands/test_import_streetscan_signs_v2.py
jforsman: True.... removed most of the stupid tests.

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
shundread: Yeah let's replace `additional-signs` with `additional_signs` then.

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
```diff
@@ -0,0 +1,2537 @@
...
+        self.sign_file = sign_file
+        self.dry_run = dry_run
+        self.force_update = force_update
+        self.delimiter = delimiter
+        self.batch_size = batch_size
+        self.user = user
+
+        # Normalise and sort by dependency order.
+        self.object_types: list[str] = [ot for ot in OBJECT_TYPE_ORDER if ot in object_types]
+        self.phases: list[str] = list(phases)
```
shundread: Do phases need sorting?

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
```diff
@@ -0,0 +1,2537 @@
...
+
+# Valid values for the --object-type and --phase CLI arguments.
+VALID_OBJECT_TYPES: tuple[str, ...] = ("mounts", "signs", "signposts", "additional-signs")
+VALID_PHASES: tuple[str, ...] = ("create", "update", "deactivate")
+
+# Dependency order — object types must be processed in this sequence.
+OBJECT_TYPE_ORDER: tuple[str, ...] = ("mounts", "signs", "signposts", "additional-signs")
+
+# Mounts are never deactivated; the deactivate phase is silently skipped for them.
+_DEACTIVATABLE_OBJECT_TYPES: frozenset[str] = frozenset({"signs", "signposts", "additional-signs"})
```
shundread, line 54: This is the second case of `frozenset` ever used in this project. The first case was also added during this feature development. I don't understand why this is done only for two specific objects and nothing else in the entire project.

* `INVALID_CODES` is a constant and a set but not a `frozenset`
* `COLOR_DEPENDENT_CODES` is a constant and a set but not a `frozenset`

You don't need to change this but we should consider trying to keep the source code in the project consistent.

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
```diff
@@ -0,0 +1,2537 @@
...
+
+        # --- MountType lookup (by Finnish and English description) ---
+        self.mount_types_by_name: dict[str, MountType] = self._build_mount_types_by_name()
+
+        # --- DB PK maps (source_id → db pk) ---
+        # Used for update and deactivate FK resolution.
+        self.mount_source_id_to_db_id: dict = self._build_mount_source_id_to_db_id()
+        self.sign_source_id_to_db_id: dict = self._build_sign_source_id_to_db_id()
+        self.additional_sign_source_id_to_db_id: dict = self._build_additional_sign_source_id_to_db_id()
+        self.signpost_source_id_to_db_id: dict = self._build_signpost_source_id_to_db_id()
```
shundread, line 173: Looking at some of the stuff that shows up later, it looks to me that the implementation of these `build_x_source_id_to_db_id` functions have a problem:

```python
    def _build_signpost_source_id_to_db_id() -> dict[str, str]: # <--------------------
        """Build mapping from source_id to db_id for SignpostReal objects.

        Returns:
            dict[str, str]: Dictionary mapping source_id to db_id.
        """
        return DbBuilderMixin._build_source_id_to_db_id(SignpostReal)
```

```python
    def _build_source_id_to_db_id(model_class: Type) -> dict[str, str]:
        """Build mapping from source_id to db_id for objects of the given model.

        Args:
            model_class (type): Django model class.

        Returns:
            dict[str, str]: Dictionary mapping source_id to db_id.
        """
        return {
            obj.source_id: obj.id # <-------- obj.id is UUID not str
            for obj in model_class.objects.filter(source_name__startswith=TrafficSignImporter.SOURCE_NAME)
        }
```

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
```diff
@@ -0,0 +1,2537 @@
...
+                is incremented.
+        """
+        phase_started_at = datetime.datetime.now(tz=datetime.timezone.utc)
+        summary.setdefault("signposts_created", 0)
+        details_before = len(summary.get("details", []))
+
+        # In-run map: source_id → db pk for signposts created during pass 1.
+        # Combined with the pre-existing signpost_source_id_to_db_id it gives
+        # the full parent-resolution map available during pass 2.
+        newly_created: dict[str, int] = {}
```
shundread: `source_id (str) -> db.pk (UUID)` but type is `dict[str, int]`?

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
```diff
@@ -0,0 +1,2537 @@
...
+
+        existing_source_ids: set[str] = set(self.signpost_source_id_to_db_id.keys())
```
shundread: I'm going to mention this one more because it hasn't been fixed in the other PR that I reviewed.

There are a log of needless casts to `set` for `some_dict.keys()` in this feature. `some_dict.keys()` is already dict-like and does everything you need for what you are using the variable for.

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
```diff
@@ -0,0 +1,2537 @@
...
+            for s, row in self.signposts_by_id.items()
+            if s not in existing_source_ids and row.get(CSVHeadersV2.status) != "Removed"
+        ]
+        # Partition into roots (no parent) and children (have parent_sign_id).
+        root_ids = [s for s in candidate_source_ids if not self.signposts_by_id[s].get(CSVHeadersV2.parent_sign_id, "")]
+        child_ids = [s for s in candidate_source_ids if s not in root_ids]
+
+        created_count = self._run_signpost_pass(
+            source_ids=root_ids,
+            parent_map={**self.signpost_source_id_to_db_id, **newly_created},
```
shundread: No need to `{**a, **b}` here, `newly_created` is guaranteed to be empty at this point.

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
```diff
@@ -0,0 +1,2537 @@
...
+            "_deactivate_signs: deactivated=%d (of %d candidates)", deactivated_count, len(deactivate_source_ids)
+        )
+        self._record_phase_result(summary, "signs", "deactivate", deactivated=deactivated_count)
+        self._save_run_log(summary)
+
+    # ------------------------------------------------------------------
+    # Signpost handlers
+    # ------------------------------------------------------------------
+
+    def _create_signposts(self, summary: dict[str, Any]) -> None:
```
shundread, line 1662: We've got an issue here where if we got an ancestry tree deeper than one parent, this will import anything deeper than that without its parents.

```
import Grandparent <- Parent <- Node <- ...

Grandparent.parent_id == None
Parent.parent_id == Grandparent.id
Node.parent_id == None
(everything else under Node).parent_id == None
```

What's the strategy for fixing that situation?

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
```diff
@@ -0,0 +1,2537 @@
...
+                {
+                    "action": "create",
+                    "object_type": "SignpostReal",
+                    "db_id": str(obj.id),
+                    "source_id": obj.source_id,
+                }
+            )
+        return len(created)
+
+    def _update_signposts(self, summary: dict[str, Any]) -> None:
```
shundread, line 1862: There's a big block of near-identical implementation between these.

Image

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
```diff
@@ -0,0 +1,2537 @@
...
+                skipped_count += 1
+                continue
+
+            new_location, device_type_id, new_mount_real_id, new_mount_type = (
+                fields["location"],
+                fields["device_type_id"],
+                fields["mount_real_id"],
+                fields["mount_type"],
+            )
+            new_location_specifier = fields["location_specifier"]
```
shundread: We have the block above

```python
            new_location, device_type_id, new_mount_real_id, new_mount_type = (
                fields["location"],
                fields["device_type_id"],
                fields["mount_real_id"],
                fields["mount_type"],
            )
```

which is functionally identical to

```python
            new_location = fields["location"]
            device_type_id = fields["device_type_id"]
            new_mount_real_id = fields["mount_real_id"]
            new_mount_type = fields["mount_type"]
```

And yet not even two lines of code apart we change from one style of writing the assignments from a `a, ..., z = (A, ..., Z)` to `a = A; ... z = Z`

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
```diff
@@ -0,0 +1,2537 @@
...
+                skipped_count += 1
+                continue
+
+            fields = self._resolve_signpost_update_fields(row, source_id, details)
+            if fields is None:
+                skipped_count += 1
+                continue
+
+            new_location, device_type_id, new_mount_real_id, new_mount_type = (
+                fields["location"],
```
shundread: `_update_additional_signs`, a near clone of this method does this instead:

```python
                new_location = self._georeferenced_point_from_csv_row(row)
```

Here we seem to just take whatever is in the CSV, even `"foobar"` Is the concern for building a valid georeferenced `Point` not relevant to signposts?

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
```diff
@@ -0,0 +1,2537 @@
...
+                fields["mount_real_id"],
+                fields["mount_type"],
+            )
+            new_location_specifier = fields["location_specifier"]
+            new_value, new_height, new_direction = fields["value"], fields["height"], fields["direction"]
+            new_condition, new_scanned_at = fields["condition"], fields["scanned_at"]
+            new_attachment_url, new_txt = fields["attachment_url"], fields["txt"]
+            new_source_name = "StreetScan2025"
+
+            if not self._signpost_fields_changed(
```
shundread: Once again with the meaningless implementation differences between nearly identical methods `_update_additional_signs` we got there just a direct comparison in-place:

```python
            internal_info = row.get("internal_additional_info")
            new_additional_information = self._build_additional_information(txt, number_code_str, internal_info)

            changed = (
                self.force_update
                or obj.source_name != new_source_name
                or obj.location != new_location
                or obj.device_type_id != device_type_id
                or obj.parent_id != new_parent_id
                or obj.signpost_real_id != new_signpost_real_id
                or obj.mount_real_id != new_mount_real_id
                or obj.mount_type_id != (new_mount_type.pk if new_mount_type else None)
                or obj.direction != new_direction
                or obj.height != new_height
                or obj.condition != new_condition
                or obj.location_specifier != new_location_specifier
                or obj.color != new_color
                or obj.additional_information != new_additional_information
                or obj.scanned_at != new_scanned_at
                or obj.attachment_url != new_attachment_url
            )
            if not changed:
                skipped_count += 1
                continue
```

Versus a method call here.

The inclusion of a method does nothing to cut down on lines of code because the parameter count is so large, plus we're introducing the opportunity to create a painfully difficult-to-debug problem by doing


```python
if not self._utility_changed(
    obj,
    A,
    B,
    C,
    E, # oops, D was skipped but good luck spotting that
    D, # oops, E was supposed to be here
    ...
)
```

versus

```python
changed = (
    obj.a != A,
    obj.b != B,
    obj.c != C,
    obj.d != E, # error here
    ...
)
```

Where the botched comparison is visible within the same line of code.

This method's approach to the same problem is the inferior one.

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
```diff
@@ -0,0 +1,2537 @@
...
+            obj.updated_by = self.user
+            obj.updated_at = phase_started_at
+            batch.append(obj)
+
+            if len(batch) >= self.batch_size:
+                self._flush_signpost_batch(batch, update_fields)
+                updated_count += len(batch)
+                batch = []
+
+        self._flush_signpost_batch(batch, update_fields)
```
shundread: In most places where you are trying to update in batches all you need to do is:

```python
updated_count += SomeModel.objects.update(batch, update_fields, batch_size=self.batch_size)
batch = []
```

It doesn't matter if `batch` is an empty array, it will simply do nothing then. These special case handles are just inflating the code for nobody's benefit.

```python
In [23]: User.objects.bulk_update([u], ["username"], batch_size=10)
Out[23]: 1

In [24]: User.objects.bulk_update([], ["username"], batch_size=10)
Out[24]: 0

In [25]: User.objects.bulk_update([u, User(username="Non existent")], ["username"], batch_size=10)
Out[25]: 1

In [26]: User.objects.get(username="Non existent")
---------------------------------------------------------------------------
DoesNotExist                              Traceback (most recent call last)
...

```

The return value is the number of objects actually affected. If an object doesn't exist in the DB it's simply skipped by the operation and it will not cause any problems.

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
```diff
@@ -0,0 +1,2537 @@
...
+
+        Args:
+            row (dict): Enriched CSV row for the signpost.
+            source_id (str): Source identifier for logging.
+            details (list[dict]): Mutable details list to append skip/warning entries.
+
+        Returns:
+            dict[str, Any] | None: Resolved field dict, or None if the row must be skipped.
+        """
+        try:
```
shundread: I missed particular detail in the criticism of `_update_signposts` not validating location because of course it needs to be done at a different spot of where it's done elsewhere. It's like this Pull Request is that bubble of the Annihilation story where everything just becomes a drifting fractal of the same pattern over and over.

This pattern keeps repeating itself nearly verbatim btw:

```
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-            MountReal: Mutated (unsaved) MountReal instance ready for bulk_update.
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-        """
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-        skipped: list[int] = summary.setdefault("_skipped_mount_update_count", [0])
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-        for source_id in update_source_ids:
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-            row = self.mounts_by_id[source_id]
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-            obj = existing.get(db_id_map[source_id])
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-            if obj is None:
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-                skipped[0] += 1
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-                continue
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-            try:
traffic_control/analyze_utils/traffic_sign_data_v2_import.py:                new_location = self._georeferenced_point_from_csv_row(row)
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-            except (KeyError, ValueError) as exc:
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-                summary.setdefault("details", []).append(
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-                    {"level": "skip", "source_id": source_id, "reason": f"Invalid coordinates on update: {exc}"}
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-                )
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-                skipped[0] += 1
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-                continue
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-            if not geometry_is_legit(new_location):
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-                summary.setdefault("details", []).append(
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-                    {
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-                        "level": "skip",
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-                        "source_id": source_id,
--
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-        """
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-        skipped: list[int] = summary.setdefault("_skipped_sign_update_count", [0])
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-        details: list[dict] = summary.setdefault("details", [])
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-        for source_id in update_source_ids:
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-            row = self.signs_by_id[source_id]
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-            obj = existing.get(db_id_map[source_id])
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-            if obj is None:
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-                skipped[0] += 1
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-                continue
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-            try:
traffic_control/analyze_utils/traffic_sign_data_v2_import.py:                new_location = self._georeferenced_point_from_csv_row(row)
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-            except (KeyError, ValueError) as exc:
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-                details.append(
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-                    {"level": "skip", "source_id": source_id, "reason": f"Invalid coordinates on update: {exc}"}
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-                )
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-                skipped[0] += 1
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-                continue
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-            if not geometry_is_legit(new_location):
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-                details.append(
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-                    {
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-                        "level": "skip",
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-                        "source_id": source_id,
--
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-        field values. Returns ``None`` and appends a skip entry to ``details``
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-        when a hard validation error is encountered.
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-        Args:
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-            row (dict): Enriched CSV row for the signpost.
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-            source_id (str): Source identifier for logging.
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-            details (list[dict]): Mutable details list to append skip/warning entries.
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-        Returns:
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-            dict[str, Any] | None: Resolved field dict, or None if the row must be skipped.
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-        """
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-        try:
traffic_control/analyze_utils/traffic_sign_data_v2_import.py:            new_location = self._georeferenced_point_from_csv_row(row)
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-        except (KeyError, ValueError) as exc:
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-            details.append({"level": "skip", "source_id": source_id, "reason": f"Invalid coordinates on update: {exc}"})
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-            return None
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-        if not geometry_is_legit(new_location):
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-            details.append(
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-                {"level": "skip", "source_id": source_id, "reason": f"Invalid location on update: {new_location.ewkt}"}
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-            )
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-            return None
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-        code = row.get(CSVHeadersV2.code, "")
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-        device_type_id = self.code_to_device_type_id.get(code)
--
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-            obj = existing.get(db_id_map[source_id])
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-            if obj is None:
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-                skipped_count += 1
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-                continue
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-            txt = row.get(CSVHeadersV2.txt, "") or ""
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-            if txt.strip().lower() == "unreadable":
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-                details.append({"level": "skip", "source_id": source_id, "reason": "text value is unreadable"})
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-                skipped_count += 1
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-                continue
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-            try:
traffic_control/analyze_utils/traffic_sign_data_v2_import.py:                new_location = self._georeferenced_point_from_csv_row(row)
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-            except (KeyError, ValueError) as exc:
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-                details.append(
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-                    {"level": "skip", "source_id": source_id, "reason": f"Invalid coordinates on update: {exc}"}
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-                )
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-                skipped_count += 1
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-                continue
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-            if not geometry_is_legit(new_location):
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-                details.append(
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-                    {
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-                        "level": "skip",
traffic_control/analyze_utils/traffic_sign_data_v2_import.py-                        "source_id": source_id,
```

The code is begging for a single way to get location.

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
shundread: It feels like there's a 
```python
for (field, value) in fields.items():
    setattr(obj, field, value)
```

pattern that would cut down a lot of this stuff considering that the set of fields to be updated are returned by `_resolve_signpost_update_fields`.

This of course applies to many parts of this file.

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
```diff
@@ -0,0 +1,2537 @@
...
+            "value": self._get_sign_value(row.get(CSVHeadersV2.number_code, "") or ""),
+            "height": self._get_sign_height(row.get(CSVHeadersV2.height)),
+            "direction": self._get_sign_direction(row.get(CSVHeadersV2.direction)),
+            "condition": self._get_sign_condition(row.get(CSVHeadersV2.condition)),
+            "scanned_at": self._get_scanned_at(row.get(CSVHeadersV2.scanned_at)),
+            "attachment_url": row.get(CSVHeadersV2.attachment_url, ""),
+            "txt": row.get(CSVHeadersV2.txt, "") or None,
+        }
+
+    def _signpost_fields_changed(  # noqa: PLR0913
```
shundread: This is not the way to do this. Discussed in a different comment.

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
```diff
@@ -0,0 +1,2537 @@
...
+            or obj.height != new_height
+            or obj.condition != new_condition
+            or obj.location_specifier != new_location_specifier
+            or obj.value != new_value
+            or obj.txt != new_txt
+            or obj.scanned_at != new_scanned_at
+            or obj.attachment_url != new_attachment_url
+        )
+
+    def _deactivate_signposts(self, summary: dict[str, Any]) -> None:
```
shundread, line 2104: Refactor this

Image

traffic_control/tests/management/commands/test_traffic_sign_importer_v2_create_mounts.py
```diff
@@ -0,0 +1,482 @@
...
+
+    Uses mock.patch to simulate the owner being missing without touching the DB,
+    since migration 0022_initial_owners seeds both owners in the test database.
+
+    Args:
+        tmp_path (Path): Pytest tmp_path fixture.
+    """
+    mf = _write_csv(tmp_path, _MOUNT_CSV_HEADER, [])
+    sf = _write_csv(tmp_path, _SIGN_CSV_HEADER, [])
+    with patch(_LOAD_OWNERS_PATH, side_effect=RuntimeError("Required Owner 'Helsingin kaupunki' not found")):
```
shundread, line 178: Why patch the method instead of ensuring the owner does not exist?

If you patch the method in the tests and the method is later updated to tolerate missing owners this test will become a lie. This is much worse than a flaky test.

Look through any tests you got that have `patch(...)` and rethink if there's a better way to test how the class will actually behave or if the test is necessary at all.

traffic_control/tests/management/commands/test_traffic_sign_importer_v2_create_mounts.py
```diff
@@ -0,0 +1,482 @@
...
+        default_owner: Owner fixture.
+    """
+    bad_row = _mount_row("SKIP_ME", x="0.0", y="0.0")
+    importer = _make_importer(tmp_path, [bad_row])
+    summary: dict = {"details": []}
+    importer._create_mounts(summary)
+
+    result = summary["phase_results"]["mounts"]["create"]
+    detail_skips = len([e for e in summary["details"] if e["level"] == "skip"])
+    assert result["skipped"] == detail_skips
```
shundread, line 405: I feel like this test should also check that the count of skipped elements is 1

traffic_control/tests/management/commands/test_traffic_sign_importer_v2_signposts.py
```diff
@@ -0,0 +1,980 @@
...
+        device_type=device_type,
+        location=Point(25497100.0, 6673400.0, 5.0, srid=3879),
+    )
+
+    importer = _make_importer(tmp_path, [_sign_row("SPUA")], phases=["update"])
+    summary: dict = {"details": []}
+    importer._update_signposts(summary)
+
+    sp = SignpostReal.objects.get(source_id="SPUA")
+    assert sp.updated_at is not None
```
shundread, line 657: Probable repeat of another bogus test. `sp.updated_at` is probably already populated when `SignpostRealFactory` creates it.

Look for tests similar to `_sets_updated_at` here and in other files and fix to compare it to a specific timestamp or confirm that before you run the update operation `updated_at` is None.

traffic_control/tests/management/commands/test_traffic_sign_importer_v2_signposts.py
```diff
@@ -0,0 +1,980 @@
...
+
+    importer = _make_importer(tmp_path, [_sign_row("SPFU", scanned_at="")], force_update=True, phases=["update"])
+    summary: dict = {"details": []}
+    importer._update_signposts(summary)
+
+    assert summary["signposts_updated"] == 1
+
+
+@pytest.mark.django_db
+def test_update_signposts_skips_removed_rows(tmp_path: Path, default_owner, device_type, mount_type) -> None:
```
shundread, line 735: These `test_update_x_skips_removed_rows` tests should also contain other row edit, refresh the object from DB and check that the edits did not go through to the object. Look for other tests like this in this file and other files and fix those too.

traffic_control/tests/management/commands/test_traffic_sign_importer_v2_signposts.py
```diff
@@ -0,0 +1,980 @@
...
+    """
+    SignpostRealFactory(
+        source_id="SPDRYU",
+        source_name="StreetScan2025",
+        owner=default_owner,
+        device_type=device_type,
+        location=Point(25497100.0, 6673400.0, 5.0, srid=3879),
+    )
+
+    importer = _make_importer(tmp_path, [_sign_row("SPDRYU")], dry_run=True, phases=["update"])
```
shundread, line 850: This test should try to object something else in the object and confirm that the field hasn't been updated. Look for similar dry_run tests that may be missing meaningful object edit data.

traffic_control/tests/management/commands/test_traffic_sign_importer_v2_update_mounts.py
```diff
@@ -0,0 +1,473 @@
...
+    summary: dict = {"details": []}
+    importer._update_mounts(summary)
+
+    existing.refresh_from_db()
+    assert existing.source_name == "StreetScan2025"
+    assert summary["mounts_updated"] == 1
+
+
+@pytest.mark.django_db
+def test_update_mounts_sets_updated_by_and_updated_at(tmp_path: Path, default_owner, mount_type) -> None:
```
shundread, line 198: The test name and docstring talks about `updated_by` but never looks at `updated_by`

shundread (changes requested):
Man you really need to hit the breaks of this feature if you ever want to get it approved.

The amount of duplicated sloppy code in this is unreal. Just because SonarCube isn't accusing it it doesn't mean it's not there. Maybe it's crashing under the sheer size of this behemoth.

A quick AI-slopped one-liner to compare how much of the entire project's python code is present in this the `v2` importer alone accuses this feature of accounting up to **15% of the entire python code of the whole project and it's still growing**. This is absurd and it's making me seriously rethink my life choices.


```shell
# move venv, .venv out of the way and run this at the project root
$ find . -type f -name "*.py" -print0 | xargs -0 wc -c | grep -v " total$" | sort -n | awk '{size[NR]=$1; path[NR]=$2; total+=$1} END { printf "%-8s | %s\n", "% Total", "File"; print "-------------------------"; for(i=1; i<=NR; i++) { pct=(total>0 ? (size[i]/total)*100 : 0); printf "%5.1f%%   | %s\n", pct, path[i] } }' | grep v2
  0.2%   | ./traffic_control/tests/management/commands/test_import_streetscan_signs_v2.py
  0.3%   | ./traffic_control/management/commands/analyze_traffic_sign_data_v2.py
  0.3%   | ./traffic_control/analyze_utils/traffic_sign_data_v2_constants.py
  0.3%   | ./traffic_control/analyze_utils/traffic_sign_data_v2.py
  0.3%   | ./traffic_control/management/commands/import_streetscan_signs_v2.py
  0.3%   | ./traffic_control/analyze_utils/traffic_sign_data_v2_data_loading.py
  0.5%   | ./traffic_control/analyze_utils/traffic_sign_data_v2_db_builders.py
  0.5%   | ./traffic_control/tests/management/commands/test_traffic_sign_importer_v2_update_mounts.py
  0.6%   | ./traffic_control/tests/management/commands/test_traffic_sign_importer_v2_create_mounts.py
  0.7%   | ./traffic_control/analyze_utils/traffic_sign_data_v2_code_transform.py
  0.8%   | ./traffic_control/analyze_utils/traffic_sign_data_v2_status_reports.py
  1.0%   | ./traffic_control/tests/management/commands/test_traffic_sign_importer_v2_signs.py
  1.1%   | ./traffic_control/tests/management/commands/test_traffic_sign_importer_v2_signposts.py
  1.3%   | ./traffic_control/tests/management/commands/test_traffic_sign_importer_v2_additional_signs.py
  1.3%   | ./traffic_control/analyze_utils/traffic_sign_data_v2_reports.py
  2.2%   | ./traffic_control/tests/management/commands/test_analyze_traffic_sign_data_v2.py
  3.9%   | ./traffic_control/analyze_utils/traffic_sign_data_v2_import.py
```

The way I see this, you got two paths forward for getting this feature in.

1. Keep piling code away until you're satisfied and just go ahead and merge this without my approval, because I will simply not approve this PR the way it's going.
2. Find yourself a nice design pattern for data imports that allows you to do stuff like:
    * `for (model, relevant_fields, row_validators, etc) in MODEL_LIST: run_create_step(model, relevant_fields, validators, etc)`
    * Fields that are not unique to a model should have a `load_FIELDNAME()` method defined in a single place
    * Some tests that are particularly repetitive should go into a single parametrized test.
    * And so on and so on, please use your extensive experience as a dev to think of what else, don't delegate this any further

What I'd really like to see from any further work into this branch is something along the lines of:

```shell
$ git diff `the-pull-request-how-it-looked-like-when-i-submitted-the-review` --shortstat
  SOME files changed, A_RELATIVELY_SMALL_NUMBER insertions(+), A_VERY_LARGE_NUMBER deletions(-)
```

traffic_control/tests/management/commands/test_traffic_sign_importer_v2_additional_signs.py
jforsman: No... they can also be parameter from @parametrize

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
jforsman: it does

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
jforsman: We dont have that kind of situation, but yeah ofc it could be possible, especially in the next scan data so better to have it supported already here. Will change from 2-pass to multi-pass I guess

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
jforsman: I think you did not mean SomeModel.objects.update but bulk_update?

jforsman: And no, not in a single place, because of the dry-run handling.

traffic_control/tests/management/commands/test_traffic_sign_importer_v2_create_mounts.py
jforsman: because owners are created with migrations, does not change the fact that someone could change it in the database. It is necessary to test that the command fails like it is supposed to.

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
jforsman: So any other problem than the annotations?

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
jforsman: All these now report a warning as was agreed with PO.

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
jforsman: Should be cleaned, and yes sonar did trigger duplicate code issues.

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
jforsman: warnings reported now for all the _get_xxx methods.

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
jforsman: Probably some leftovers when I had trouble getting the attachment files actually written to blobstorage, will check it again.

jforsman: Ah... yeah... it is because the temp file is opened with delete=False -> needs to be manually deleted.  I would not go touching this anymore.

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
jforsman: warning as the rest of these.

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
jforsman: warning as the rest of these.

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
jforsman: warning as the rest of these.

traffic_control/admin/streetscan_import.py
```diff
@@ -0,0 +1,574 @@
...
+            obj (StreetScanImportRevertFile): The revert file instance.
+
+        Returns:
+            str: HTML anchor to the parent run's change page, or dash if unsaved.
+        """
+        if not obj.pk or not obj.import_run_id:
+            return "-"
+        url = reverse("admin:traffic_control_streetscanimportrun_change", args=[obj.import_run_id])
+        return mark_safe(
+            render_to_string(
```
shundread: `render_to_string` does not need to be wrapped by `mark_safe`.

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
```diff
@@ -0,0 +1,2472 @@
...
+            obj = build_object(source_id, row, location, details, phase_started_at)
+            if obj is None:
+                continue
+            processed.append(source_id)
+            yield obj
+
+    def _create_objects(
+        self,
+        summary: dict[str, Any],
+        rows_by_id: dict[str, Any],
```
shundread, line 410: Nitpick: Id put a `*` in the args list either after `self` or after `summary` parameter to force arguments to be passed by name. You are already using it in that way, but with four `str` positional parameters in a row it's easy to get arguments mixed up unless you're passing arguments by name.

Putting `*` after `summary` would require no changes anywhere else.

traffic_control/analyze_utils/traffic_sign_data_v2_import.py
```diff
@@ -0,0 +1,2472 @@
...
+        Returns:
+            int: Number of objects in the batch (same whether dry run or not).
+        """
+        if not self.dry_run:
+            model_class.objects.bulk_update(batch, update_fields, batch_size=self.batch_size)
+        return len(batch)
+
+    def _update_objects(
+        self,
+        summary: dict[str, Any],
```
shundread, line 876: Consider also forcing these parameters to be passed by name since we have 3 `dict[str, Any]` in a row in positional parameters. If you put `*` after `summary` no changes will be needed in how this method is already used.

shundread (commented):
This is all the feedback I can offer on short notice. Any further feedback would require me to look at the entire PR as a whole again and that's not happening.

shundread (approved):
"""Approved""" but really merge this at your own peril

PR #865, opened 2026-06-01
chore: Tweak logging of import_operational_areas_hki_wfs command

shundread:
Refs: LIIK-537

### DIFF
--- a/traffic_control/tests/management/commands/test_import_operational_areas_hki_wfs.py
+++ b/traffic_control/tests/management/commands/test_import_operational_areas_hki_wfs.py
@@ -1,12 +1,31 @@
 from datetime import date
+from io import StringIO
 from unittest.mock import patch
 
 import pytest
 from django.core.management import call_command
 
 from traffic_control.models import OperationalArea
+from traffic_control.tests.factories import OperationalAreaFactory
 from traffic_control.tests.management.commands.utils import create_mock_data_source, MockAttribute, MockFeature
 from traffic_control.tests.test_base_api import test_multi_polygon
+
+MOCK_FEATURE_0 = MockFeature(
+    {
+        "id": MockAttribute(0),
+        "nimi": MockAttribute("feature 0 (updated)"),
+        "nimi_lyhyt": MockAttribute("f0"),
+        "urakkamuoto": MockAttribute("area type 0"),
+        "urakoitsija": MockAttribute("contractor 0"),
+        "alku_pvm": MockAttribute("2019-01-01"),
+        "loppu_pvm": MockAttribute("2019-12-31"),
+        "paivitetty_tietopalveluun": MockAttribute("2019-08-01"),
+        "tehtavakokonaisuus": MockAttribute("KATU"),
+        "status": MockAttribute("ready"),
+    },
+    test_multi_polygon.ogr,
+)
+
 
 MOCK_FEATURE_1 = MockFeature(
     {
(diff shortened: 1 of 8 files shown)


traffic_control/tests/management/commands/test_import_operational_areas_hki_wfs.py
```diff
@@ -43,21 +61,45 @@
 
 @pytest.mark.django_db
 def test_importer_created_operational_areas():
+    OperationalArea.objects.create(
```
jforsman: Add a factor class for OperationArea (we even have or had a ticket that all tests should use factoryboy classes) and use that instead.

traffic_control/tests/management/commands/test_import_operational_areas_hki_wfs.py
```diff
@@ -43,21 +61,45 @@
...
+        status="ready",
+        source_name="urakkarajat_katu",
+        location=" ".join(
+            [
+                line.strip()
+                for line in """
+            SRID=3879;MULTIPOLYGON Z (((25497733.5 6672927.5 0, 25497946.5
+            6673032.5 0, 25498653.5 6673034.5 0, 25498987.5 6672708.5 0, 25498314.5
+            6672170.5 0, 25497651.5 6672629.5 0, 25497646.5 6672775.5 0, 25497733.5
+            6672927.5 0)))""".split("\n")
```
jforsman: Where is this coming from? I think you could use some already defined polygons from traffic_control/tests/test_base_api.py traffic_control/tests/test_base_api_3d.py. Or even move those to some utility file for tests if you like.

traffic_control/tests/management/commands/test_import_operational_areas_hki_wfs.py
```diff
@@ -43,21 +61,45 @@
...
+        new=create_mock_data_source([MOCK_FEATURE_0, MOCK_FEATURE_1, MOCK_FEATURE_2]),
     )
+    out = StringIO()
 
     with patch_url, patch_ds:
-        call_command("import_operational_areas_hki_wfs")
+        call_command("import_operational_areas_hki_wfs", stdout=out)
 
-    assert OperationalArea.objects.count() == 1
+    assert OperationalArea.objects.count() == 2
```
jforsman: I don't get that why this is == 2 and not == 3? There should be 3 objects right? Even the output asserts starting from line 113 checks that 2 synced from source and 1 newly created?

jforsman: Ohwell.. MockFeature2 has  "tehtavakokonaisuus": MockAttribute("OTHER-TASK"), so that is left out. 
So something wrong with the command output? as it states 2 synced and 1 created.

jforsman (changes requested):

traffic_control/tests/management/commands/test_import_operational_areas_hki_wfs.py
shundread: Fixed

traffic_control/tests/management/commands/test_import_operational_areas_hki_wfs.py
shundread: Fixed, I had a sloppy approach to doing this which was searching for MULTIPOLYGON in the source code and copying a commonly reoccurring example here. It's now just using the factory-provided value.

traffic_control/tests/management/commands/test_import_operational_areas_hki_wfs.py
shundread: Fixed the string.

jforsman (approved):
Looks good now. Also great that now you removed the get_operational_area function totally.

PR #873, opened 2026-06-08
fix: devicetype code mapping to database id was not done correctly

jforsman:
* Was possible that device type code was mapped to the wrong dbid with the legacy code

Refs: LIIK-977


shundread (approved):
This change looks fine to me.

PR #877, opened 2026-06-11
fix: 403 response when using openlayers tile provider

jforsman:
* Start using Helsinki's own tile service

Refs: LIIK-953


shundread (changes requested):
Doesn't work for signposts apparently. Not sure if it's broken for other objects too

shundread (commented):
works now

shundread (approved):
works now

PR #880, opened 2026-06-16
Nplus one fixes

jforsman:


shundread (approved):
Looks good. Neat find on the guardian n+1 hit

PR #885, opened 2026-06-17
fix: avoid N+1 DB queries in post_files bulk file upload

jforsman:
* self.get_file_relation was called within loop that caused extra db hits
* cache related parent so it is not unnecessarily fetched multiple times

Refs: LIIK-982

### DIFF
--- a/traffic_control/serializers/common.py
+++ b/traffic_control/serializers/common.py
@@ -3,7 +3,7 @@ from uuid import UUID
 
 from django.core.exceptions import ValidationError
 from django.utils.translation import gettext_lazy as _
-from drf_spectacular.utils import extend_schema_field
+from drf_spectacular.utils import extend_schema_field, extend_schema_serializer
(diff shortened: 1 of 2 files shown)


traffic_control/serializers/common.py
```diff
@@ -74,8 +74,85 @@ class EwktGeometryField(serializers.CharField):
     pass
 
 
+class _CachedParentRelatedField(serializers.PrimaryKeyRelatedField):
```
shundread: The implementation looks more generic than a "parent-specific" `_CachedParentRelatedField`. Why private something that can serve a more general purpose?

shundread (approved):
Passes tests, clears 28 warnings in the file uploads tests. Looks good to me (with one comment)

traffic_control/serializers/common.py
jforsman: True... changed the name.

PR #886, opened 2026-06-22
fix: nplusone problems in embed view

jforsman:
Refs: LIIK-983

### DIFF
--- a/traffic_control/views/embed.py
+++ b/traffic_control/views/embed.py
@@ -6,9 +6,106 @@ from django.views.generic import DetailView
 
 from traffic_control.models import MountPlan, MountReal, TrafficSignPlan, TrafficSignReal
 
+# ---------------------------------------------------------------------------
+# Shared select_related fragments
+# ---------------------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


shundread (approved):
This is a pretty weird fix after all the parent abstract class ends up with needless baggage that never gets used because it is necessarily overwritten.

Consider writing the TrafficSignPlanEmbed / TrafficSignRealEmbed in terms of of `get_queryset` specializations?

```python
class TraficSignEmbed:
    def get_queryset():
        stuff().select_related(common_fields)

class TrafficSignPlanEmbed
    def get_queryset():
        super().get_queryset().select_related(plan_only_fields)
```

Something like that could also be done for get_additional_signs by splitting and specializing the part that gets the queryset for additional signs.

If too much work, at least get rid of the completely unused `additional_sign_select_related` and `traffic_sign_select_related`

alternatively declare the overriding fields in terms of what's defined in the base class

`foo_select_related = BaseClass.foo_select_related + (specialized_sleect_related_fields...)`

Finally

* it's three common fields in traffic_sign_select_related, not two
* it's four common fields in additional_sign_select_related, not three

PR #888, opened 2026-06-26
fix: n+1 problems with replace feature

jforsman:
Refs: LIIK-989

### DIFF
--- a/traffic_control/signal_utils.py
+++ b/traffic_control/signal_utils.py
@@ -1,5 +1,6 @@
 import logging
 import os
+from typing import Any, Optional
 
 import cairosvg
 from auditlog.models import LogEntry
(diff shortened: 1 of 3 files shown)


traffic_control/signal_utils.py
```diff
@@ -92,6 +98,85 @@ def _create_parent_log_entry(parent, message, action=None):
...
+    original_from_db = model.from_db
+
+    @classmethod  # type: ignore[misc]
+    def patched_from_db(cls, db: str, field_names: list, values: list) -> Any:
+        instance = original_from_db.__func__(cls, db, field_names, values)
+        setattr(instance, loaded_attr, instance.__dict__.get(source_id_attr))
+        return instance
+
+    model.from_db = patched_from_db
+    model._signal_utils_patched_from_db_attrs = already_patched | {loaded_attr}
```
shundread: Nitpick: Since we're using `getattr` + `_PATCHED_FROM_DB_ATTRS_KEY` to check for "already patched" case, I would use `setattr` + `_PATCHED_FROM_DB_ATTRS_KEY` here just to guard from the possibility that `_PATCHED_FROM_DB_ATTRS_KEY` ever becomes a different string in some future commit.

shundread (approved):
It took me a couple of reads to understand it, as I haven't done much caching in Django. Nice to see dozens of n+1 warnings disappear. One nitpick.

PR #896, opened 2026-08-05
feat: Add app with admin view to track usage of management commands

shundread:
To track management command usage, you need to update the command's base class and initiate tracking in the admin panel for the app.You can visit the admin list view for the app for instructions on how to do so.

To configure which apps are scanned for trackable commands you can configure the TRACK_COMMAND_USAGE_APPS value in the project's settings. Otherwise the apps list is assumed to be cityinfra, city_furniture, traffic_control, users.

Refs: LIIK-966


command_tracker/admin.py
```diff
@@ -0,0 +1,139 @@
...
+def is_trackable(app_name: str, cmd_name: str) -> bool:
+    try:
+        cmd_instance = load_command_class(app_name, cmd_name)
+        return isinstance(cmd_instance, TrackableCommand)
+    except Exception:
+        return False
+
+
+def get_tracked_apps():
+    """Helper to evaluate the setting at runtime so tests can override it."""
```
jforsman: Well... tests can also override settings. Not saying that this should be removed, just a comment about the comment.

command_tracker/admin.py
```diff
@@ -0,0 +1,139 @@
...
+            ),
+            path(
+                "<path:object_id>/delete/",
+                self.admin_site.admin_view(self.delete_view),
+                name=f"{info[0]}_{info[1]}_delete",
+            ),
+        ]
+
+    def custom_changelist_view(self, request):
+        existing_commands = {k: v for k, v in get_commands().items() if v in get_tracked_apps()}
```
jforsman: for clarity k,v could be named differently, or atleast comment the sematics of those.

command_tracker/admin.py
```diff
@@ -0,0 +1,139 @@
...
+        tracked_command_entries = {obj.id: obj for obj in TrackedManagementCommand.objects.all()}
+
+        ctx_tracked_commands = []
```
jforsman: having ctx prefix makes it harder to read maybe use postfix instead?

command_tracker/tests.py
```diff
@@ -0,0 +1,260 @@
...
+        commands_dict format: {"command_name": {"trackable": bool, "exists": bool}}
+        """
+        # Clean up any previously started patches if _setup is called multiple times
+        for p in active_patches:
+            p.stop()
+        active_patches.clear()
+
+        # Mock `get_commands` to return only the commands configured as existing
+        active_cmds = {k: "track_example_app" for k, v in commands_dict.items() if v["exists"]}
+        print("active_cmds is", active_cmds)
```
jforsman: I think better policy is to use logging instead of prints even in the tests.

command_tracker/models.py
```diff
@@ -0,0 +1,35 @@
...
+    Tracking the usage of some commands might help reveal whether some of them have gone without use for a very long
+    time and are candidates for deletion from the codebase.
+    """
+
+    id = models.CharField(primary_key=True, max_length=201, editable=False)  # Formatted as 'app_name-command_name'
+    app = models.CharField(max_length=100, editable=False)
+    command = models.CharField(max_length=100, editable=False)
+    comment = models.TextField(
+        blank=True,
+        null=True,
```
jforsman: for text fields it would be better not to have blank=True and null=True, better just to have blank=True.

jforsman (changes requested):

command_tracker/admin.py
shundread: This comment is wrong and was a result of iteration with Gemini. I earlier had something like `TRACKED_APPS = the return expression of this function` at the root of the file, and THAT doesn't get updated by tests settings overrides.

The value needs to be evaluated at use time, not module load time to work properly with pytest settinigs overrides. The AI just documented this wrong.

command_tracker/admin.py
shundread: Fixed

command_tracker/admin.py
shundread: Fixed

command_tracker/tests.py
shundread: Leftover debug print, removed

command_tracker/models.py
shundread: Fixed

jforsman (approved):
looks good.

PR #901, opened 2026-08-06
fix: Split tracking start and execution timestamps on command tracker

shundread:
Refs: LIIK-966


jforsman (approved):
translations missing, it is ok just to have the .po files updated with the msgids and empty translations. Otherwise looks ok.

PR #907, opened 2026-08-10
fix: uv lock for release please

jforsman:
* fix application version not updated to uv.lock file

Refs: LIIK-993


shundread (approved):
seems legit

PR #910, opened 2026-08-12
feat: AI translations

jforsman:
* Missing translation to fi and sv po-files added by AI
* added a script that uses Anthropic API to translate missing translations
* Script needs API KEY, so it is there just for a referece
* Maintenance guide in confluence is also updated with instructions

Refs: LIIK-898


shundread (approved):
Looks ok

PR #911, opened 2026-08-12
fix: More idiomatic error responses for plan bulk insertion endpoint

shundread:
Refs: LIIK-967

### DIFF
--- a/traffic_control/serializers/bulk_plan_insert.py
+++ b/traffic_control/serializers/bulk_plan_insert.py
@@ -3,6 +3,7 @@ from collections import defaultdict
 
 from django.db import transaction
 from rest_framework import serializers
+from rest_framework.settings import api_settings
 
 from traffic_control.serializers.additional_sign import (
     AdditionalSignPlanInputSerializer,
@@ -12,6 +13,7 @@ from traffic_control.serializers.mount import MountPlanInputSerializer, MountPla
 from traffic_control.serializers.plan import PlanSerializer
 from traffic_control.serializers.signpost import SignpostPlanInputSerializer, SignpostPlanOutputSerializer
 from traffic_control.serializers.traffic_sign import TrafficSignPlanInputSerializer, TrafficSignPlanOutputSerializer
+from traffic_control.serializers.utils import get_single_object_serializer
 
 BULK_PLAN_INSERT_MOCK_BATCH_PAYLOAD = {
     "additional_sign_plans": [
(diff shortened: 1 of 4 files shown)


traffic_control/tests/test_plan_bulk_insert_api.py
```diff
@@ -168,7 +171,11 @@ def _post_insert_plan_bulk(
     payload = json.dumps(payload_obj, indent=2, default=str)
 
     print(f"Request is:\n{payload}\n\n")
-    return admin_client.post(reverse("v1:plan-bulk-insert"), data=payload, content_type="application/json")
+    response = admin_client.post(reverse("v1:plan-bulk-insert"), data=payload, content_type="application/json")
+    if response.status_code != status.HTTP_201_CREATED:
+        response_data = response.json()
+        print(f"Error information is:\n{json.dumps(response_data, indent=2)}\n\n")
```
jforsman: print used here

jforsman: And the whole if statement is useless anyways?

traffic_control/serializers/bulk_plan_insert.py
```diff
@@ -224,14 +283,19 @@ class BulkPlanInputResponseSerializer(serializers.Serializer):
 
 
 def to_list(value) -> list:
```
jforsman: Could have error handling or support for other types, like dict or any other iterable (from collections.abc import Iterable).

traffic_control/views/bulk_plan_insert.py
```diff
@@ -224,14 +283,19 @@ class BulkPlanInputResponseSerializer(serializers.Serializer):
...
+def get_single_object_serializer(serializer_class: serializers.ModelSerializer) -> serializers.ModelSerializer:
+    """Resolve the child serializer of a list field, or the main serializer for single object field."""
     if hasattr(serializer_class, "child"):
-        return serializer_class.child, True
-    return serializer_class, False
+        return serializer_class.child
+    return serializer_class
+
+
+def is_array_serializer(serializer_class: serializers.ModelSerializer) -> bool:
```
jforsman: is the type annotation correct? Should it also include ListSerializer? I think DRF internally actually creates ListSerializer instead of ModelSerializer if called with many=True.

traffic_control/serializers/bulk_plan_insert.py
```diff
@@ -224,14 +283,19 @@ class BulkPlanInputResponseSerializer(serializers.Serializer):
...
+    """Cast the value as a list if it is a single object."""
     if not isinstance(value, list):
         return [value]
     return value
```
jforsman: All the free functions should be moved to some other file (eg. views._common.py), just to keep the view file clean. And those are generic functions, not specific to this view.

jforsman (changes requested):
Also all serializers should be move the same location as where the other view serializers are: traffic_control/serializers have separate files for different objects.

Did not find any function issues.

traffic_control/serializers/bulk_plan_insert.py
shundread: We don't need it for this feature. Should I just `_private` it?

jforsman: Sure.

jforsman (approved):
Just fix the to_list and make it private, then good to go.

PR #915, opened 2026-08-14
fix: Add missing validation to location fields

shundread:
Refs: LIIK-998

### DIFF
--- a/traffic_control/serializers/common.py
+++ b/traffic_control/serializers/common.py
@@ -1,6 +1,7 @@
 from typing import Optional, OrderedDict
 from uuid import UUID
 
+from django.contrib.gis.geos import GEOSException, GEOSGeometry
 from django.core.exceptions import ValidationError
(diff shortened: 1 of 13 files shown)


shundread:
Hold off on the review. I think I need to revise this.

traffic_control/serializers/common.py
```diff
@@ -23,7 +41,9 @@
     }
 )
 class EwktPointField(serializers.CharField):
-    pass
+    def to_internal_value(self, data):
+        internal_data = super().to_internal_value(data)
```
jforsman, line 45: If you wanna put fancy pants on, you could make a baseclass inherited from CharField and then have this to_internal_value there. eg.

class BaseGeomField(serializer.CharField):
      geom_types = []

      def to_internal_value(self, data):
            internal_data = super().to_internal_value(data)
            return parse_and_validate_geometry_field(internal_data, self.geom_types)

jforsman (approved):

PR #918, opened 2026-08-19
feat: Management command to set additional sign real parents based on…

jforsman:
… mount

* Finds trafficsignreal or signpostreal attached to the same mount as the additional sign
* ticket machines are excluded.
* Only parentless additional sign reals with mount_real are handled
* Incase of multiple matches found, the one above additional sign is selected.

Refs: LIIK-1000


traffic_control/models/link_additional_sign_parents_run_info.py
```diff
@@ -0,0 +1,62 @@
...
+
+    This model stores information about each run of the command including:
+    - Start and end times
+    - Whether it was a dry-run
+    - Which AdditionalSignReal instances were linked to which TrafficSignReal or SignpostReal parents
+    - Statistics about the run (total processed, skipped, etc.)
+    """
+
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    start_time = models.DateTimeField()
```
shundread: `started_at` is a more common name for the field indicated when the run was started in our management command `SomeCommandRun` models.

traffic_control/models/link_additional_sign_parents_run_info.py
```diff
@@ -0,0 +1,62 @@
...
+    end_time = models.DateTimeField(null=True, blank=True)
```
shundread: `completed_at` is a more common name for the field indicated when the run was completed in our management command `SomeCommandRun` models.

traffic_control/management/commands/link_additional_sign_parents_by_mount.py
```diff
@@ -0,0 +1,352 @@
...
+            self.style.WARNING(
+                f"No active TrafficSignReal above AdditionalSignReal {ads.id} with mount_real {ads.mount_real_id}"
+            )
+        )
+
+        signposts = SignpostReal.objects.active().filter(mount_real=ads.mount_real)
+        if signposts.exists():
+            signpost = self._select_best_candidate(signposts, ads, "SignpostReal", stats)
+            if signpost:
+                self.stdout.write(
```
shundread: Is this console log a bit redundant considering that `_select_best_candidate` already prints something similar?

```python
        if selected:
            self.stdout.write(
                f"Selected {model_name} {selected.id} for AdditionalSignReal {ads.id} "
                f"on mount {ads.mount_real_id} (z={selected.location.z:.2f}, z-diff={min_z_diff:.2f})"
            )
```

traffic_control/tests/management/commands/test_link_additional_sign_parents_by_mount.py
```diff
@@ -0,0 +1,1122 @@
...
+    def test_command_excludes_ticket_machines(self, ticket_machine_device_type, traffic_sign_device_type):
+        """Test that ticket machines are excluded from processing.
+
+        Args:
+            ticket_machine_device_type: Ticket machine device type fixture
+            traffic_sign_device_type: Traffic sign device type fixture
+        """
+        # Create mount, traffic sign, and ticket machine
+        mount = MountRealFactory()
+        _traffic_sign = TrafficSignRealFactory(device_type=traffic_sign_device_type, mount_real=mount)
```
shundread: It would be good if this test otherwise carries all the information that would lead to a successful link should the device type had been something else so we get a clearer confirmation that the `device_type` was the reason this wasn't linked.

So make the initial conditions a bit more like the conditions in `test_command_links_additional_sign_to_traffic_sign` but keep the `device_type` altered:

```diff
-        _traffic_sign = TrafficSignRealFactory(device_type=traffic_sign_device_type, mount_real=mount)
+        _traffic_sign = TrafficSignRealFactory(
+            device_type=traffic_sign_device_type,
+            mount_real=mount,
+            location=Point(MIN_X + 10.0, MIN_Y + 10.0, 2.0, srid=settings.SRID),
+        )
         ticket_machine = AdditionalSignRealFactory(
             device_type=ticket_machine_device_type,
             mount_real=mount,
             parent=None,
+            location=Point(MIN_X + 10.0, MIN_Y + 10.0, 2.0, srid=settings.SRID),
         )
```

traffic_control/tests/management/commands/test_link_additional_sign_parents_by_mount.py
```diff
@@ -0,0 +1,1122 @@
...
+
+        # Run command
+        out = StringIO()
+        call_command("link_additional_sign_parents_by_mount", stdout=out)
+
+        # Verify ticket machine parent was NOT set
+        ticket_machine.refresh_from_db()
+        assert ticket_machine.parent is None
+
+    def test_command_skips_additional_sign_without_mount_real(
```
shundread, line 187: Similar suggestion about the starting conditions looking a bit more like `test_command_links_additional_sign_to_traffic_sign` here.

traffic_control/tests/management/commands/test_link_additional_sign_parents_by_mount.py
```diff
@@ -0,0 +1,1122 @@
...
+    ):
+        """Test that additional signs with existing parents are skipped.
+
+        Args:
+            additional_sign_device_type: Additional sign device type fixture
+            traffic_sign_device_type: Traffic sign device type fixture
+        """
+        # Create mount with traffic signs
+        mount = MountRealFactory()
+        traffic_sign_1 = TrafficSignRealFactory(device_type=traffic_sign_device_type, mount_real=mount)
```
shundread: Should `_traffic_sign_2` be populated with data that would make it a better candidate if `traffic_sign_1` wasn't already assigned to it?

traffic_control/tests/management/commands/test_link_additional_sign_parents_by_mount.py
```diff
@@ -0,0 +1,1122 @@
...
+        """Test that a TrafficSignReal is preferred when both a traffic sign and signpost share the mount.
+
+        Args:
+            additional_sign_device_type: Additional sign device type fixture
+            traffic_sign_device_type: Traffic sign device type fixture
+            signpost_device_type: Signpost device type fixture
+        """
+        # Create mount that has both a traffic sign and a signpost above
+        mount = MountRealFactory()
+        traffic_sign = TrafficSignRealFactory(
```
shundread, line 467: could put the traffic sign higher than the signpost so the test shows that the reason the traffic sign was picked was the device type, not closer height to the additional sign.

traffic_control/tests/management/commands/test_link_additional_sign_parents_by_mount.py
```diff
@@ -0,0 +1,1122 @@
...
+
+        # Run command
+        out = StringIO()
+        call_command("link_additional_sign_parents_by_mount", stdout=out)
+
+        # Verify the active one was selected (soft-deleted excluded)
+        ads.refresh_from_db()
+        assert ads.parent == active_traffic_sign
+
+    def test_command_selects_latest_active_traffic_sign(self, additional_sign_device_type, traffic_sign_device_type):
```
shundread, line 697: The name of the test and what is actually being tested seem to differ.

The name implies that it will show the logic that when you have two or more equally acceptable traffic_sign candidates to link to (same height), it will link to the latest added.

But it isn't much different to `test_command_handles_multiple_traffic_signs_same_mount` mixed with `test_command_prefers_active_traffic_sign_over_newer_soft_deleted`

Put all traffic signs on same height?

traffic_control/tests/management/commands/test_link_additional_sign_parents_by_mount.py
```diff
@@ -0,0 +1,1122 @@
...
+
+        # Run command
+        out = StringIO()
+        call_command("link_additional_sign_parents_by_mount", stdout=out)
+
+        # Verify closest active was selected
+        ads.refresh_from_db()
+        assert ads.parent == closest_active
+
+    def test_command_selects_latest_soft_deleted_traffic_sign_when_no_active(
```
shundread: Put traffic signs on same height

traffic_control/tests/management/commands/test_link_additional_sign_parents_by_mount.py
```diff
@@ -0,0 +1,1122 @@
...
+        # Run command
+        out = StringIO()
+        call_command("link_additional_sign_parents_by_mount", stdout=out)
+
+        # Verify active signpost was selected
+        ads.refresh_from_db()
+        assert ads.parent is None
+        assert ads.signpost_real == active_signpost
+
+    def test_command_selects_latest_soft_deleted_signpost_when_no_active(
```
shundread, line 836: If the height is not a factor in the decision, make the latest look like a less appealing option in relation to height

traffic_control/tests/management/commands/test_link_additional_sign_parents_by_mount.py
```diff
@@ -0,0 +1,1122 @@
...
+
+        # Verify appropriate skip reason - no active candidates at all
+        run_info = LinkAdditionalSignParentsRunInfo.objects.latest("start_time")
+        assert run_info.skipped_no_match == 1
+        assert len(run_info.skipped_records) == 1
+        assert run_info.skipped_records[0]["additional_sign_real_id"] == str(ads.id)
+        assert run_info.skipped_records[0]["mount_real_id"] == str(mount.id)
+        assert run_info.skipped_records[0]["reason"] == "no_matching_active_traffic_sign_or_signpost"
+
+    def test_command_excludes_ticket_machines_from_signpost_fallback(
```
shundread, line 886: Similar comment as in `test_command_excludes_ticket_machines`

shundread (changes requested):
Implementation looks fine. Many tests seem to have initial conditions that make it unclear to see which behavior in the command exactly triggered the result to be what it is.

For example

```python
def test_that_command_ignores_inactive_thing():
    ad = AdditionalSignRealFactory(..., z=1.0)
    s1 = Thing(..., z=2.0, is_active=True)
    s2 = Thing(
        ....,
        is_active=False,  # possible reason for s1 to be preferred over this
        z=3.0, # another possible reason for s1 to be preferred over this
    )
```

It's possible that `test_command_tie_breaking_by_created_at_for_same_z` becomes redundant if some of the suggestions here are applied.

traffic_control/management/commands/link_additional_sign_parents_by_mount.py
jforsman: Yes it is redundant, will remove from here.

traffic_control/tests/management/commands/test_link_additional_sign_parents_by_mount.py
jforsman: sure

traffic_control/tests/management/commands/test_link_additional_sign_parents_by_mount.py
jforsman: Well.. the name of the test function is bogus, if there are only softdeletes, nothing should be selected, like the docstring states. So the height does not matter.

traffic_control/tests/management/commands/test_link_additional_sign_parents_by_mount.py
jforsman: Name of the function was misleading, docstring was correct, will be change to 
`test_command_selects_nothing_when_no_active_traffic_signs`

shundread (approved):
lgtm

PR #921, opened 2026-08-21
Fix/liik 995 send zero users deactivated emails

shundread:
Also added a small HTML fix for email  table using deprecated HTML properties.


jforsman (approved):
Looks good to me.

